PR #9, opened 2021-01-11
#8 Convert codebase to TypeScript

### LINKED ISSUE #8: Package is not compatible with CommonJS 'require()'
I'm having trouble using the package on a NodeJS backend project.

For the frontend, ESM is ok as it gets converted by Babel which is usually in the setup.
However, regarding 'vanilla' backend NodeJS project, I'm having issue because I can't `require()` this lib:

image
image

```
Error [ERR_REQUIRE_ESM]: Must use import to load ES Module: /Users/test-import/node_modules/erc725.js/src/index.js
require() of ES modules is not supported.
require() of /Users/test-import/node_modules/erc725.js/src/index.js from /Users/test-import/index.js is an ES module file as it is a .js file whose nearest parent package.json contains "type": "module" which defines all .js files in that package scope as ES modules.
Instead rename /Users/test-import/node_modules/erc725.js/src/index.js to end in .cjs, change the requiring code to use import(), or remove "type": "module" from /Users/test-import/node_modules/erc725.js/package.json.
```

In order to make it work, people have to set up an extra babel or rollup pipeline to convert it. This could reduce adoption.

I would suggest to make it CJS compatible. Also, as this would likely imply adding an extra build phase in the project, I would recommend converting the project to TypeScript as we could get both types and CommonJS target.

I can send a draft PR with TS support and also GitHub Actions for CI/CD to npm.

Hugoo:
Closes #8 

- Converted codebase from plain js to TypeScript
- Added as much types as possible but left some `any` and `@ts-ignore`
- Added ERC725 related interfaces and export them so users can use the interface in their code (`/src/types`)
- TypeScript is compiled into `commonjs` so the lib can now be used in a "classic" nodejs repo
- Added `prepare` and `prepublishOnly` scripts and `.npmignore`
- Added GitHub Actions script to perform lint and test on push (we can later add auto deploy to npm) (need to check if it works)
- converted functions comments to JSDoc syntax

PR goes to main, should I change to `develop` ?

## Examples in a TS project:

image
image
image
image


src/lib/encoder.ts
```diff
@@ -201,62 +253,62 @@ export const valueContentEncodingMap = {
...
-        return (value)
-            ? valueTypeEncodingMap[type].decode(value)
-            : value
+        return value ? valueTypeEncodingMap[type].decode(value) : value
 
     },
 
-
-    encodeValueContent: (type, value) => {
+    encodeValueContent: (type: string, value: string): string | false => {
```
Hugoo, line 283: @robertdavid010 is this output type valid? The code returns either a `string` or `false`. I'd prefer not to mix types if possible.

src/lib/encoder.ts
```diff
@@ -201,62 +253,62 @@ export const valueContentEncodingMap = {
...
+            return type === value ? value : false
 
         }
 
         return value ? valueContentEncodingMap[type].encode(value) : '0x'
 
     },
 
-    decodeValueContent: (type, value) => {
+    decodeValueContent: (type: string, value: string): string | false => {
```
Hugoo, line 299: same here

PR #12, opened 2021-02-10
Update module exports, graphql queries, documentation.

robertdavid010:
And minor improvements.


.npmignore
```diff
@@ -2,4 +2,5 @@ tsconfig.json
 src
 test
 .idea
-.vscode
\ No newline at end of file
+.vscode
+.github
```
Hugoo, line 6: If we use GitHub actions, we want to keep this folder. I know GitHub actions are not working now so maybe you can simply delete the folder instead of putting it in Gitignore?

src/lib/queries.ts
```diff
@@ -68,7 +69,7 @@ export const queries = {
     {
       erc725(id:"${id}") {
         id
-        dataStore
+        # dataStore
```
robertdavid010, line 72: This is a yet to be properly defined relationship in the model... since I don't see this highly used I didn't add it yet.

src/index.ts
```diff
@@ -50,28 +50,29 @@ export class ERC725 {
...
-      } // TODO: Add check for schema format
+      }
 
       // Init options member
       this.options = {
           schema,
           address,
           providerType: null,
-          ipfsGateway: 'https://cloudflare-ipfs.com/ipfs/' // 'https://ipfs.infura-ipfs.io/ipfs/'
+          ipfsGateway: ipfsGateway || 'https://ipfs.lukso.network/ipfs/' // 'https://cloudflare-ipfs.com/ipfs/' // 'https://ipfs.infura-ipfs.io/ipfs/'
```
Hugoo, line 75: What about putting the default value in the constructor arguments?

```
constructor(schema: Erc725Schema[], address?: string, provider?: any, ipfsGateway = 'https://ipfs.lukso.network/ipfs/')
```

?

What about putting this variable in a global variable / config file? (not sure if relevant tho)

Hugoo (commented):
Side note:

I noticed the `getAllData()` function is not typed / has wrong type:

image

So when using it I need to add `: any`

image

We can probably improve the types here and maybe add generics maybe.

robertdavid010:
> Side note:
> 
> I noticed the `getAllData()` function is not typed / has wrong type:
> 
> image
> 
> So when using it I need to add `: any`
> 
> image
> 
> We can probably improve the types here and maybe add generics maybe.

Yeah this is probably impossible to strictly type, since we cant know what data format will be used by user 100%.

rryter:
@robertdavid010 What's the status with this PR, should I review or is it done? As of right now we are using the lib in the Universal Profile Website, and we should switch the default IPFS gateway to ours.

src/index.ts
```diff
@@ -50,28 +50,29 @@ export class ERC725 {
...
-    address;
-    providerType: ProviderType | null;
+    address?;
+    providerType?: ProviderType | null;
     provider?;
     ipfsGateway: string;
   };
 
-  constructor(schema: Erc725Schema[], address?: string, provider?: any) {
+  constructor(schema: Erc725Schema[], address?: string, provider?: any, ipfsGateway?: string) {
```
frozeman, line 59: the last object should be an options object.

src/index.ts
```diff
@@ -50,28 +50,29 @@ export class ERC725 {
...
-      } // TODO: Add check for schema format
+      }
 
       // Init options member
       this.options = {
           schema,
           address,
           providerType: null,
-          ipfsGateway: 'https://cloudflare-ipfs.com/ipfs/' // 'https://ipfs.infura-ipfs.io/ipfs/'
+          ipfsGateway: ipfsGateway || 'https://cloudflare-ipfs.com/ipfs/' // 'https://ipfs.lukso.network/ipfs/' // 'https://ipfs.infura-ipfs.io/ipfs/'
```
frozeman, line 75: `ipfsGateway: options ? options.ipfsGateway || 'https://cloudflare-ipfs.com/ipfs/'`

frozeman (changes requested):
Documentation is missing

src/index.ts
robertdavid010: Interesting idea... for next PR.

robertdavid010:
> Side note:
> 
> I noticed the `getAllData()` function is not typed / has wrong type:
> 
> image
> 
> So when using it I need to add `: any`
> 
> image
> 
> We can probably improve the types here and maybe add generics maybe.

Challenging to explicitly define types, as we can not control the schema being used. Maybe the user could provide the types with the schemas. Also good for a next PR.

> Documentation is missing

Documentation is now updated

src/index.ts
robertdavid010: ipfs gateway is handled as a string currently... we could update this in the future(?)

robertdavid010: This is now in a separate issue.

src/index.ts
robertdavid010: we can connect this with the params object updates issue?

PR #15, opened 2021-05-20
Develop

robertdavid010:


docs/source/content/getting-started.rst
```diff
@@ -33,25 +33,63 @@ Or alternately the named export in cases where it assists with integration betwe
 
 
 --------------------------------------------------
-Instantiation
+Schema
```
frozeman, line 36: that is not a schema, but an example, so it needs to be clear that this is just an example

frozeman: This should be in the example section of the instantiation, not up here

frozeman (changes requested):

docs/source/content/getting-started.rst
robertdavid010: Updated

PR #17, opened 2021-06-07
chore(axios): add custom log message on error

rryter:
When a request made via **axios** fails, log the error and also print out the URL the request was sent to. This should simplify spotting a potential URL encoding problem especially when running a script via node.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -5,7 +5,7 @@
     "declaration": true,
     "outDir": "./lib",
     "strict": true,
-    "noImplicitAny": false,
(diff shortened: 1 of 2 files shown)


tsconfig.json
```diff
@@ -5,7 +5,7 @@
     "declaration": true,
     "outDir": "./lib",
     "strict": true,
-    "noImplicitAny": false,
+    "noImplicitAny": false, //TODO: this should be removed
```
Hugoo, line 8: 👍

Hugoo (commented):
lgtm

PR #18, opened 2021-06-09
Copy Docs content to JSDoc

Hugoo:
Content created for the docs is now available as JSDoc.
This enables IDE to show documentation in the code.

image

I also added a few more TS types


rryter:
Looks good to me. @robertdavid010 how do we proceed?

PR #19, opened 2021-06-14
Optimize Build

rryter:
- provide two builds (commonjs & esm)
- removed unnecessary dependencies
- replaced @apollo/client with graphql-tag
- updated a few imports to allow treeshaking later on
- added mocha config file
- resolved all the issues of "npm audit"

### DIFF
--- a/src/lib/constants.ts
+++ b/src/lib/constants.ts
@@ -1,6 +1,4 @@
-// put all constants here
-import * as Web3Utils from 'web3-utils'
-
+import { numberToHex, keccak256 } from 'web3-utils'
 import { MethodData, Encoding, Method } from '../types/Method'
 
 interface Constants {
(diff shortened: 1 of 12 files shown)


src/lib/constants.ts
```diff
@@ -1,6 +1,4 @@
-// put all constants here
-import * as Web3Utils from 'web3-utils'
-
+import { numberToHex, keccak256 } from 'web3-utils'
```
Hugoo, line 1: 👍

tsconfig.json
```diff
@@ -1,14 +1,49 @@
...
+    "noUnusedLocals": true /* Report errors on unused locals. */,
+    "noUnusedParameters": true /* Report errors on unused parameters. */,
+    "noImplicitReturns": true /* Report error when not all code paths in function return a value. */,
+    "noFallthroughCasesInSwitch": true /* Report errors for fallthrough cases in switch statement. */,
+
+    /* Debugging Options */
+    "traceResolution": false /* Report module resolution log messages. */,
+    "listEmittedFiles": false /* Print names of generated files part of the compilation. */,
+    "listFiles": false /* Print names of files part of the compilation. */,
+    "pretty": true /* Stylize errors and messages using color and context. */,
```
Hugoo, line 36: 👍

tsconfig.json
```diff
@@ -1,14 +1,49 @@
...
+
+    // "strict": true /* Enable all strict type-checking options. */,
+
+    /* Strict Type-Checking Options */
+    // "noImplicitAny": true /* Raise error on expressions and declarations with an implied 'any' type. */,
+    // "strictNullChecks": true /* Enable strict null checks. */,
+    // "strictFunctionTypes": true /* Enable strict checking of function types. */,
+    // "strictPropertyInitialization": true /* Enable strict checking of property initialization in classes. */,
+    // "noImplicitThis": true /* Raise error on 'this' expressions with an implied 'any' type. */,
+    // "alwaysStrict": true /* Parse in strict mode and emit "use strict" for each source file. */,
```
Hugoo, line 24: Do you want to keep these comments?

tsconfig.json
```diff
@@ -1,14 +1,49 @@
...
+
+    /* Debugging Options */
+    "traceResolution": false /* Report module resolution log messages. */,
+    "listEmittedFiles": false /* Print names of generated files part of the compilation. */,
+    "listFiles": false /* Print names of files part of the compilation. */,
+    "pretty": true /* Stylize errors and messages using color and context. */,
+
+    /* Experimental Options */
+    // "experimentalDecorators": true /* Enables experimental support for ES7 decorators. */,
+    // "emitDecoratorMetadata": true /* Enables experimental support for emitting type metadata for decorators. */,
```
Hugoo, line 40: same here?

tsconfig.json
```diff
@@ -1,14 +1,49 @@
...
+
+    "lib": ["es2017", "dom"],
+    "types": [],
+    "typeRoots": ["node_modules/@types", "src/types"]
   },
-  "include": ["src"],
-  "exclude": ["node_modules", "tests/*"]
+  "include": ["src/**/*.ts"],
```
Hugoo, line 46: Is it necessary to add `*.ts` only? I don't know if it adds "more value" and it may be confusing if later we had .js / .json files in the repo and then it doesn't compile well because of this rule. I would prefer to keep `src` only.

tsconfig.module.json
```diff
@@ -0,0 +1,12 @@
+{
+    "extends": "./tsconfig",
+    "compilerOptions": {
+      "target": "esnext",
+      "outDir": "build/module",
+      "module": "esnext"
+    },
+    "exclude": [
+      "node_modules/**"
+    ]
```
Hugoo, line 10: so it means for this build you want to re-add the `tests/*` ? Why ?
Is it necessary to add this exclude in this tsconfig file?

Hugoo (commented):
Cool !

tsconfig.json
rryter: No, I will remove them.

tsconfig.json
rryter: same here :)

tsconfig.json
rryter: I agree, I will revert to `src` only.

tsconfig.module.json
rryter: Good find, this was not intentional.

rryter:
Adjusted as per your feedback @Hugoo. Thanks a lot!

PR #20, opened 2021-06-25
Adjust configuration and extend tree-shaking support

rryter:
For review purposes it's best to look at the commits individually.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,7 +53,7 @@ const mySchema = [
   },
 ];
 
-let myERC725 = new ERC725(schema, address, provider[, config])
+let myERC725 = new ERC725(schema[, address, provider, config])
 ```
 
 ## Usage
--- a/README.md
+++ b/README.md
@@ -53,7 +53,7 @@ const mySchema = [
   },
 ];
 
-let myERC725 = new ERC725(schema[, address, provider, ipfsGateway])
+let myERC725 = new ERC725(schema, address, provider[, config])
 ```
 
 ## Usage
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -53,7 +53,7 @@ const mySchema = [
   },
 ];
 
-let myERC725 = new ERC725(schema[, address, provider, ipfsGateway])
```
frozeman, line 56: i guess we want to add here an example config object and also use the variables from above.

best move line 28 and 27 below 54 you

src/index.ts
```diff
@@ -24,7 +24,9 @@ import axios from 'axios'
 import GraphSource from './providers/subgraphProviderWrapper'
 import Web3Source from './providers/web3ProviderWrapper'
 import EthereumSource from './providers/ethereumProviderWrapper'
-import { utils } from './lib/utils'
+import {
+    decodeAllData, decodeKey, decodeKeyValue, encodeAllData, encodeArrayKey, encodeKey, getSchemaElement
+} from './lib/utils'
```
frozeman, line 29: Isn't the libs just an internal library? I think selecting here functions doesnt improve a lot. External dependencies are the problem

frozeman (changes requested):

src/index.ts
rryter: I assumed the utils would also be imported directly from time to time. Either way it doesnt hurt imo.

I'm setting up a little project to investigate the size properly.

README.md
rryter: Yes, very good input.

PR #21, opened 2021-06-28
Drop Axios dependency and use ES2015 for the module build instead of ESNEXT

rryter:
- Fixed the module build
- Excluded docs from the npm package
- Dropped Axios dependency


rryter:
New size:
```
npm notice package size:  46.0 kB                                 
npm notice unpacked size: 242.8 kB  
```

.github/workflows/node.js.yml
```diff
@@ -0,0 +1,30 @@
...
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Use Node.js ${{ matrix.node-version }}
+      uses: actions/setup-node@v2
+      with:
+        node-version: ${{ matrix.node-version }}
+    - run: npm ci
+    - run: npm run build --if-present
+    - run: npm test
```
Hugoo, line 30: 👏  super cool !

package.json
```diff
@@ -42,6 +42,7 @@
   "devDependencies": {
     "@types/mocha": "^8.2.2",
     "@types/node": "^15.12.2",
+    "@types/graphql": "^14.5.0",
```
Hugoo, line 45: Why do we need these types if we remove `graphql` from the dep?
image

test/test.ts
```diff
@@ -31,6 +31,8 @@ import {
     generateAllData, generateAllRawData, generateAllResults
 } from './testHelpers'
 
+require('isomorphic-fetch')
```
Hugoo, line 34: Can you do "import" instead?

```
import 'isomorphic-fetch'
```

Hugoo (commented):
I think having `;` at the eol makes the code much readable.
What about activating this rule in prettier/linter ?

package.json
```diff
@@ -51,13 +52,11 @@
     "mocha": "^9.0.0",
     "npm-run-all": "^4.1.5",
     "ts-node": "^10.0.0",
-    "typescript": "^4.3.2"
+    "typescript": "^4.3.2",
+    "isomorphic-fetch": "^3.0.0"
```
Hugoo, line 56: Maybe you want to also install the types for this dep? https://www.npmjs.com/package/@types/isomorphic-fetch

```
npm install --save-dev @types/isomorphic-fetch
```

Also, is this a dev dependancy? I would assume it isn't but i might be wrong.

package.json
rryter: I don't know for sure yet. I think it's related to the `graphql-tag` dependency. I could not get the github actions to run successfully without them. I'm currently investigating.

package.json
rryter: It actually *is* a dev dependency. Moch runs the tests in a node environment without `fetch` available. It's only necessary because 3 tests actually perform real `fetch` requests. We might as well mock `fetch` that would probably be cleaner.

test/test.ts
rryter: Yeah, I'll update this.

rryter:
Regarding the formatting, I totally agree. We should offload the formatting to prettier 100%. I'll prepare a PR for that.

@frozeman any objections to that?

frozeman (approved):

Issue #16, opened 2021-06-04
Allow user to encode json. Currently he must hash json first and then encode the hashed json.


antoniopgs:

robertdavid010:
This is currently beyond the scope of the featureset, and is a categorical extension of said featureset. I'm not sure this is appropriate to do in this library, as there are other libraries which would this.

rryter:
I think this could actually be a helpful feature, let's discuss this @robertdavid010 and check if we can include it in some upcoming changes. Or would you @antoniopgs be willing to provide a PR?

There is also already a todo in the code for this:
https://github.com/ERC725Alliance/erc725.js/blob/main/src/lib/encoder.ts#L226

frozeman:
I agree, we actually already hash it when using `.fetchData()` here: https://github.com/ERC725Alliance/erc725.js/blob/087e9bcb5a5445b2c017aaf0aec2f5828f6a1aa3/src/index.ts#L408
, so we can easily encode JSON files with the hash. 

@rryter can you add this to your TODO pls?

something like:
```js
erc725.encodeData('LSP3Profile', {
    json: myJson,
    url: 'https://myURL.com/jsonfile.json'
 })
// coud return
0x6f357c6a828dc0b16e548a1dc32a3730413cf06b52ab87ac877c72bc50c4500f2d71d23e697066733a2f2f516d664c3753616a6f35486345327570556538596b6b3177576a58535535654b5733454d35707a444a4b456e506e

```

antoniopgs:
@frozeman why not reduce it to one argument only? either:
- pass in the json, and the ipfs cid is calculated in the background
- pass in the cid, and the json gets processed in the background

why make the user provide two inputs which can be derived from each other?

rryter:
@antoniopgs People can upload their files wherever they want, it's not guaranteed to be on IPFS.
@frozeman It's added on my TODO.

PR #27, opened 2021-07-15
Test implementation for #16 and small coverage increase

rryter:
- Slightly rearranged test setup
- added test for #16


src/lib/utils.ts
```diff
@@ -402,25 +402,19 @@ export function getSchemaElement(schemas: Erc725Schema[], key: string) {
...
       return keccak256(JSON.stringify(data));
 
     case 'keccak256(bytes)':
+    case '0x8019f9b1':
       return keccak256(data);
 
     default:
       throw new Error(
-        `Chosen hashFunction ${hashFunction.toLowerCase()} is not supported`,
+        `Chosen hashFunction '${hashFunction.toLowerCase()}' is not supported`,
```
Hugoo, line 417: I would suggest to do:

```ts
enum SUPPORTED_HASH_FUNCTIONS {
  KECCAK256_UTF8 = 'keccak256(utf8)',
  HASH_KECCAK256_UTF8 = '0x6f357c6a',
  KECCAK256_BYTES = 'keccak256(bytes)',
  HASH_KECCAK256_BYTES = '0x8019f9b1',
}

export function hashData(data: unknown, hashFunction: SUPPORTED_HASH_FUNCTIONS)
```

+ use enum in switch

+ in error, add the list of supported hash functions:

```
`hashFunction '${hashFunction.toLowerCase()}' is not supported. Supported hashFunction: ${listValuesOfEnum(SUPPORTED_HASH_FUNCTIONS)}`
```

May be "overkill" but I think it adds clarity and makes things cleaner.

Here, hashFunctions is a finite set of strings, so it would make more sense for me to us an enum rather than more global "string" ?

The values from this enum can also be used elsewhere in the code, cf. other comment.

src/index.test.ts
```diff
@@ -20,29 +20,73 @@
...
+        // @ts-ignore
+        // eslint-disable-next-line no-new
+        new ERC725(mockSchema, address, { test: false });
+      },
+      (error) =>
+        error.message.indexOf('Incorrect or unsupported provider') >= -1,
+    );
+  });
+
+  it('should throw when incorrect calling getData without address & provider options set', async () => {
```
Hugoo, line 74: i don't think the word "incorrect" adds value here. Why not simply:

"should throw when calling getData without address & provider options set"

Hugoo (commented):
nice improvements ! 👍

src/lib/encoder.test.ts
```diff
@@ -0,0 +1,50 @@
...
+      url: 'https://file-desination.com/file-name',
+      json: {
+        test: true,
+      },
+    });
+    assert.deepStrictEqual(
+      result,
+      '0x6f357c6a35709f35a87bedba632da6c0c99c927a532a27aee80f24556a7ba06723dd3c0168747470733a2f2f66696c652d646573696e6174696f6e2e636f6d2f66696c652d6e616d65',
+    );
+  });
```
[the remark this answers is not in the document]
rryter: very good input! i updated it accordingly

src/lib/utils.ts
rryter: Good idea, I changed this accordingly

PR #34, opened 2021-08-15
Refactoring:  Providers             

### LINKED ISSUE #32: Refactoring (Web3|Ethereum|Graph)Source
As per discussion with @frozeman I suggest to do the following:
- [x] Rename `(Web3|Ethereum|Graph)Source` to `(Web3|Ethereum|Graph)ProviderWrapper`
- [x] Do not include `GraphProviderWrapper` in `index.ts` to avoid the dependency to `graphQL` when not actually using said provider
- [x] Update Docs accordingly
- [x] Write / Update Tests

rryter:
- **What kind of change does this PR introduce?**
  - Enhancement

- **What is the current behavior?** 
  - Currently `graphql` needs to be installed alongside of `@erc725/erc725.js` because of the dependency to `graphql-tags`. 

- **What is the new behavior (if this is a feature change)?**
  - The Idea is to refactor how the `providers` are handled and also rename them from `(*)Source` to `(*)ProviderWrapper`. A new package called `@erc725/provider-wrappers` will be introduced, to hold the code for the `GraphProviderWrapper` and potential additional wrappers. The `(Ethereum|Web3)ProviderWrapper` will remain within `@erc725/erc725.js`
  - Closes #32


src/lib/provider-wrapper-utils.ts
```diff
@@ -0,0 +1,42 @@
...
+let idCount = 0;
+// @ts-ignore
+const web3abiDecoder = abi.default;
+
+export function decodeResult(method: Method, result) {
+  const rpcResult = result.result;
+  return rpcResult === '0x'
+    ? null
+    : web3abiDecoder.decodeParameter(METHODS[method].returnEncoding, rpcResult);
+}
```
Hugoo, line 15: This lib is typed so I don't understand we are not getting the types.

I tried locally and this seems to preserve the types:

image

Not sure tho if it has the same behaviour as idk what `abi.default` returns

src/index.ts
```diff
@@ -107,50 +100,37 @@ export class ERC725<Schema extends GenericSchema> {
...
-      this.options.providerType = ProviderType.WEB3;
-      this.options.provider = new Web3Source(givenProvider);
-
-      // CASE: Unknown provider
-    } else {
-      throw new Error(`Incorrect or unsupported provider ${givenProvider}`);
+      default:
+        throw new Error(
+          `Incorrect or unsupported provider ${providerOrProviderWrapper}`,
+        );
```
Hugoo, line 133: I'm not super fan and used to switch on a constant value such as `switch(true)`. Why not use if/if-else/else ?

src/providers/web3ProviderWrapper.ts
```diff
@@ -22,51 +22,43 @@
...
-
-export default class Web3Source {
-  public provider: any;
-
-  constructor(provider) {
+import { constructJSONRPC, decodeResult } from '../lib/provider-wrapper-utils';
+import { ProviderTypes } from '../types/provider';
+
+export class Web3ProviderWrapper {
+  type: ProviderTypes;
```
Hugoo, line 31: Is this part of the code getting "picked" by the typedoc? Would it make sense to add TypeDoc here and in the functions of the class so it goes into the docs?

src/lib/provider-wrapper-utils.ts
rryter: I have already tried this. I'm afraid this does not work, it does compile, but It then fails at runtime with:

```bash
TypeError: web3_eth_abi_1.AbiCoder is not a constructor
```

Maybe we can look into it together?

The package actually exports an instance, so we are not supposed to call it with `new`. When trying to use it as an instance, compilation fails.

src/index.ts
rryter: I think it's not very readable (else if) but how about this?
```typescript  
    if (typeof providerOrProviderWrapper.request === 'function')
      return new EthereumProviderWrapper(providerOrProviderWrapper);

    if (
      !providerOrProviderWrapper.request &&
      typeof providerOrProviderWrapper.send === 'function'
    )
      return new Web3ProviderWrapper(providerOrProviderWrapper);

    if (providerOrProviderWrapper.type === ProviderTypes.GRAPH_QL)
      return providerOrProviderWrapper;

    throw new Error(
      `Incorrect or unsupported provider ${providerOrProviderWrapper}`,
    );
```

src/providers/web3ProviderWrapper.ts
rryter: Yes, everything in `src` is getting picked up. However Fabian wanted to keep the docs small, and since we already have the `providers.md` I was thinking that will be sufficient. I'm cool with documenting this too, it's not a big effort.

rryter: I'll document this when I unify the way providers are injected, potentially in the next few PRs

src/providers/baseProviderWrapper.ts
```diff
@@ -1,16 +0,0 @@
...
-  provider: any;
-
-  constructor(provider) {
-    this.provider = provider;
-  }
-
-  abstract getAllData(address: string, keys: string[]): Promise<any>;
-  abstract getData(address: string, keyHash: string): Promise<any>;
-  abstract getOwner(address: string): Promise<any>;
-}
```
Hugoo: @rryter what was the reason for removing the base class and not using extends?

I'd like to introduce it again so all our providers have same functions/signatures.
Atm, `getAllData` of Graph provider has a different signature than web3|ethereum providers.

image

PR #41, opened 2021-10-11
Improve ci

Hugoo:
### What kind of change does this PR introduce? (Bug fix, feature, docs update, ...)

Improve CI: make prettier work within ESLint, apply prettier everywhere, add lint in CI, remove action on push (keep only on PR)

### What is the current behavior? (You can also link to an open issue here)

- CI runs twice (for push and for PR)
- CI does not test for lint
- Lint will not fail on format error

### What is the new behavior (if this is a feature change)?

- All problem above are fixed


.github/workflows/lint-test-build.yml
```diff
@@ -1,15 +1,12 @@
...
+name: Lint, Build and Test
 
 on:
-  push:
-    branches:
-      - '*'
   pull_request:
     branches:
-      - '*'
+      - 'main'
```
CallumGrindle, line 9: May still be good to have this run on all pull requests? Or at least also on PR to `develop`

PR #58, opened 2021-11-18
fix: handling of missing keys

### LINKED ISSUE #53: Requesting an array key which doesn't exist in the schema breaks
### I'm submitting a...

- bug report

### Summary

This profile: [`0x4b0b32DEF87EB60C0cAE3E3c2B4F23e3B5fd8DA3`](https://erc725-inspect.lukso.tech/?address=0x4b0b32DEF87EB60C0cAE3E3c2B4F23e3B5fd8DA3) has not this key set:

```js
{
    name: "LSP3IssuedAssets[]",
    key: "0x3a47ab5bd3a594c3a8995f8fa58d0876c96819ca4516bd76100c92462f2f9dc0",
    keyType: "Array",
    valueContent: "Address",
    valueType: "address",
}
```

When trying to fetch this key with ERC725, it throws and breaks `getData()`:

image

### Expected behavior

`getData()` function should not break if a key is missing, it should return `undefined` and still allow to get the other keys

-> Check also the behavior of `fetchData` and other key types. It broke on an array key but it may broke for just any missing key.

Hugoo:

Fix

If user requests a key which is not set in the ERC725Y contract, the function will throw

It will return '' or [] and won't throw

### Other information:

Closes #53

### DIFF
--- a/src/index.test.ts
+++ b/src/index.test.ts
@@ -106,6 +106,50 @@ describe('Running @erc725/erc725.js tests...', () => {
     const ERC725_CONTRACT_ADDRESS =
       '0x320e678bEb3369702EA14555a74414B2C531c510';
 
+    it('should return empty if the key does not exist in the contract', async () => {
+      const erc725 = new ERC725(
+        [
+          {
+            name: 'ThisKeyDoesNotExist',
+            key: '0xb12a0af5f83066646eb63c96bf29dcb827024d9a33189f5a61652a03951d1fbe',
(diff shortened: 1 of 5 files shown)


CJ42:
Nice and useful behaviour for interfaces! 🙂

frozeman:
Please return null or []

Hugoo:
> Please return null or []

I updated the PR, see last commit e2b7ddc

PR #67, opened 2021-11-25
perf: use one RPC call on getData multiple

### LINKED ISSUE #49: Use only one RPC call for `getAllData()`
### Summary

If lib wants to request multiple data keys: A,B, C
it will perform multiple: `getData()` calls:

- For legacy: `getData(A)`, `getData(B)`, `getData(C)`
- For new: `getData([A])`, `getData([B])`, `getData([C])`

It should instead send only one request:

- For legacy: `getDataMultiple([A, B, C])`
- For new: `getData([A, B, C])`

Hugoo:

Performances

See issue #49 

Will use only 1 RPC call when possible

### Other information:

Closes #49 

Screenshot 2021-11-24 at 09 50 05

### DIFF
--- a/src/providers/ethereumProviderWrapper.ts
+++ b/src/providers/ethereumProviderWrapper.ts
@@ -26,6 +26,7 @@ import * as abi from 'web3-eth-abi';
 
 import { ERC725_VERSION, INTERFACE_IDS, METHODS } from '../lib/constants';
 import { decodeResult as decodeResultUtils } from '../lib/provider-wrapper-utils';
+import { JsonRpcEthereumProvider } from '../types/JsonRpc';
 import { Method } from '../types/Method';
 import { ProviderTypes } from '../types/provider';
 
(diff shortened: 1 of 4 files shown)


test/mockProviders.ts
```diff
@@ -181,18 +199,27 @@ export class EthereumProvider {
         {
           const keyParam = '0x' + payload.params[0].data.substr(10);
 
-          result = Array.isArray(this.returnData)
-            ? this.returnData.find((e) => e.key === keyParam).value
-            : this.returnData;
+          result = this.returnData.find((e) => e.key === keyParam).value;
```
rryter, line 202: can `this.returnData.find((e) => e.key === keyParam)` ever return undefined?

rryter (approved):

CallumGrindle (approved):

src/providers/ethereumProviderWrapper.ts
```diff
@@ -125,21 +130,65 @@ export class EthereumProviderWrapper {
...
+      case ERC725_VERSION.ERC725:
+        return this.getAllDataNonLegacy(address, keys);
+      case ERC725_VERSION.ERC725_LEGACY:
+        return this.getAllDataLegacy(address, keys);
+      default:
+        return [];
+    }
+  }
+
+  private async getAllDataNonLegacy(
```
frozeman, line 152: what a horrible name :)

`getData` should be it and if there is OLD function you call it `getDataLegacy`

frozeman:
for keytype array we can only do it in 2 calls, as we first need to get the length.. Is that considered?

src/providers/ethereumProviderWrapper.ts
Hugoo: There is already a function named getData, it was the one used in previous code as there was no distinction between legacy and no legacy. I didn't know what else I could use as getData was already taken.... If I call this one getData, how do I call the other one then?

Hugoo:
> for keytype array we can only do it in 2 calls, as we first need to get the length.. Is that considered?

Yes this is handled in another part of the code, this update won't affect this.

test/mockProviders.ts
Hugoo: Good catch, yes it potentially can.
So far we only use it with a set of data we know will always be true as the mock input and "expected return data" come from same object. But adding a `?` can't hurt, I'll update thanks.

src/providers/ethereumProviderWrapper.ts
Hugoo: I'm thinking maybe the current "getData" function can be removed at least for new ERC725Y as the command is the same, only the number of items in the array is different. I will check this tomorrow and make it more "lean". There might be some useless logic duplication

frozeman (approved):

PR #69, opened 2021-12-09
docs: remove deprecated key `SupportedStandards:ERC725Account`

CJ42:

Remove deprecated key `SupportedStandards:ERC725Account`.

The code sample use an old deprecated key in the schema and does not return anything for the key `SupportedStandards:`. 

- [x] replaced deprecated key `SupportedStandards:ERC725Account` by new key `SupportedStandards:LSP3UniversalProfile`

Changed in:
- README
- md docs page
- js examples

### Other information:

- [x] added schema with `SupportedStandards:ERC725Account` as legacy schema (schema used in _universalprofiles.cloud_

### DIFF
--- a/src/index.test.ts
+++ b/src/index.test.ts
@@ -104,7 +104,7 @@ describe('Running @erc725/erc725.js tests...', () => {
     const LEGACY_ERC725_CONTRACT_ADDRESS =
       '0xb8E120e7e5EAe7bfA629Db5CEFfA69C834F74e99';
     const ERC725_CONTRACT_ADDRESS =
-      '0x320e678bEb3369702EA14555a74414B2C531c510';
+      '0x6464C9b995DA466a6fcD50C79D7e9FDd84c74b92';
 
     it('should return null if the key does not exist in the contract', async () => {
       const erc725 = new ERC725(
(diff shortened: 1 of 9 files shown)


src/index.test.ts
```diff
@@ -150,7 +150,7 @@ describe('Running @erc725/erc725.js tests...', () => {
       assert.deepStrictEqual(dataArray, { 'NonExistingArray[]': [] });
     });
 
-    const e2eSchema: any = [
+    const legacyE2eSchema: any = [
```
Hugoo, line 153: This test here is about checking whether or not the lib returns what it is supposed to return with legacy or not legacy mock.

If the schema itself is using legacy "keys" or not is not in the scope of this test.

We just check with one schema, see the results, validate or invalidate the tests. Using key A or key B is out of scope.

I therefore recommend to not duplicate the schema and not add `legacy` in the schema name.

Could you remove any occurrence of `legacy` in this test file and just keep the test

Hugoo (changes requested):
nice to update these stuff

CJ42:
@Hugoo I have made the changes on the test you requested

src/index.test.ts
```diff
@@ -192,7 +192,7 @@ describe('Running @erc725/erc725.js tests...', () => {
         web3.currentProvider,
       );
       const result = await erc725.getData();
-      assert.deepStrictEqual(result, e2eResults);
+      assert.notDeepStrictEqual(result, e2eResults);
```
Hugoo, line 195: why switch from toEqual to **not** equal ??

CJ42: @Hugoo understood the issue.
I have reverted the changes to use the same schema for the e2e tests.

PR #70, opened 2021-12-16
ci: change CI release workflow

Hugoo:

Improve CI

NPM publish is triggered from a "published GitHub release"

NPM publish can only come from `main` branch.

-> The docs are also automatically updated when the version is bumped with `npm run release`


CallumGrindle (approved):
Cool! Thinking maybe we could change the `Bump version` action name to something like `Create Release PR` as it also will update the CHANGELOG as well as bump version + create tags.

As it abstracts the whole flow of creating + pushing tags away from the developer it should be less error prone (e.g forgetting to push tags) so would be cool to use this as the primary way of making releases

Hugoo:
@CallumGrindle good point I will update it. Yeah I think the dispatch workflow with the option to put standard-release arguments is cool so there is no manual tasks at all.

Issue #78, opened 2022-01-04
Add a function `decodePermissions` to decode KM permissions


Hugoo:
### I'm submitting a...

- feature request

### Summary

Think and implement a feature which would allow users to easily decode the permissions for a KeyManager (XOR, etc.).

```js
myErc725.decodePermissions('0x0000000000000000000000000000000000000000000000000000000000000010');
/**
{
CHANGEOWNER: false,
CHANGEPERMISSIONS: false,
ADDPERMISSIONS: false,
SETDATA: false,
CALL: true,
STATICCALL: false,
DELEGATECALL: false,
DEPLOY: false,
TRANSFERVALUE: false,
SIGN: false,
}
*/
```

## References

- <https://docs.lukso.tech/standards/universal-profile/lsp6-key-manager#-address-permissions>
- #77

Hugoo:
Done in https://github.com/ERC725Alliance/erc725.js/pull/84

PR #84, opened 2022-01-05
feat: LSP6 Permissions encoding methods

### LINKED ISSUE #77: Add function `encodePermissions` to encode KM permissions
### I'm submitting a...

- feature request

### Summary

Think and implement a feature which would allow users to easily encode the permissions for a KeyManager (XOR, etc.).

```js
myErc725.encodePermissions({
CHANGEOWNER: true,
CHANGEPERMISSIONS: true,
ADDPERMISSIONS: true,
SETDATA: true,
CALL: true,
STATICCALL: true,
DELEGATECALL: true,
DEPLOY: true,
TRANSFERVALUE: true,
SIGN: true,
})
/**
0x00000000...2ff
*/
```

If one permission is not set - it defaults to `false`.

Here we can discuss this function "API".

## References

- <https://docs.lukso.tech/standards/universal-profile/lsp6-key-manager#-address-permissions>

CallumGrindle:
feature

Adds methods to encode and decode LSP6 permissions to/from hexadecimal string

Closes #77 + #78


Hugoo (approved):
Super cool !

src/index.ts
```diff
@@ -53,7 +53,12 @@ export {
 };
 
 export { ERC725Config, KeyValuePair, ProviderTypes } from './types';
-export { flattenEncodedData, encodeData } from './lib/utils';
+export {
+  flattenEncodedData,
+  encodeData,
+  encodePermissions,
+  decodePermission,
+} from './lib/utils';
```
Hugoo, line 61: I suggest to add it in the class - so instead of export from - you can import from

And then add a static function within the class.

PR #108, opened 2022-02-11
feat: isValidSignature

### LINKED ISSUE #104: Add `.isValidSignature()` method
### I'm submitting a...

- feature request

### Summary

```js
await myERC725.isValidSignature("Message or hash", signature);
```

If the input parameter is hex + 32 bytes long (char length == 66 AND starts with or without 0x) -> then we assume the input is a hash -> no pre hash before check.
If not, the lib should hash the input first.

Then, it calls `.isValidSignature()` on the contract used at init.
If we get the magic value `0x1626ba7e` -> returns true
If not -> returns false

If call error -> throw error which says "Does the address ${init contract address} support EIP-1271?"

Because the lib can be init with asset or non UP account SC addresses

### Other information

- <https://eips.ethereum.org/EIPS/eip-1271>
- Add documentation accordingly

Hugoo:
Closes #104 

## Docs page

image

image

### DIFF
--- a/src/lib/isValidSignature.ts
+++ b/src/lib/isValidSignature.ts
@@ -1,0 +1,69 @@
+/*
+    This file is part of @erc725/erc725.js.
+    @erc725/erc725.js is free software: you can redistribute it and/or modify
+    it under the terms of the GNU Lesser General Public License as published by
+    the Free Software Foundation, either version 3 of the License, or
+    (at your option) any later version.
+    @erc725/erc725.js is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU Lesser General Public License for more details.
+    You should have received a copy of the GNU Lesser General Public License
+    along with @erc725/erc725.js.  If not, see <http://www.gnu.org/licenses/>.
+*/
+/**
+ * @file lib/isValidSignature.ts
+ * @author Hugo Masclet <@Hugoo>
+ * @date 2022
+ */
+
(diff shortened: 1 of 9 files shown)


frozeman (approved):

frozeman:
@Hugoo when done?

Hugoo:
will continue working on it by eow - tmrw Fri. 4th

src/lib/isValidSignature.ts
```diff
@@ -0,0 +1,69 @@
...
+ * @param {string} messageOrHash
+ * @param {string} signature
+ * @param {string} address the contract address
+ * @returns {Promise<boolean>} Return true if the signature is valid (if the contract returns the magic value), false if not.
+ */
+export const isValidSignature = async (
+  messageOrHash: string,
+  signature: string,
+  address: string,
+  wrappedProvider: EthereumProviderWrapper | Web3ProviderWrapper, // ProviderWrapper
```
CallumGrindle, line 44: Is ProviderWrapper an unused type? Can it be removed?

CallumGrindle (approved):

test/mockProviders.ts
```diff
@@ -164,14 +188,17 @@ interface EthereumProviderPayload {
 export class EthereumProvider {
   public returnData;
   public supportsInterfaces: string[];
+  public isValidSignature: boolean;
 
   constructor(
     props: { returnData: { key: string; value: string }[] },
     supportsInterfaces: string[],
+    isValidSignature?: boolean, // to mock isValidSignature call. If set to tru, will return magic value.
```
CallumGrindle, line 196: typo `true`

CallumGrindle, line 196: typo `true`

frozeman (approved):

PR #109, opened 2022-02-18
fix: wait until a promise is resolved when using an ethereum provider

Esanim:
Ethereum provider doesn't work when trying to get the profile information.
Ethereum provider works for getting the profile information.

### DIFF
--- a/src/providers/ethereumProviderWrapper.test.ts
+++ b/src/providers/ethereumProviderWrapper.test.ts
@@ -37,5 +37,29 @@ describe('EthereumProviderWrapper', () => {
         assert.deepStrictEqual(error.message, 'some error');
       }
     });
+
+    it('should return a getData result', async () => {
+      const ethResults = [
+        '0x000000000000000000000000a78e0e7c9b1b36f7e25c5ccdfdba005ec37eadf4',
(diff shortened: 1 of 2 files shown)


src/providers/ethereumProviderWrapper.test.ts
```diff
@@ -37,5 +37,29 @@ describe('EthereumProviderWrapper', () => {
         assert.deepStrictEqual(error.message, 'some error');
       }
     });
+
+    it('should return a getData result', async () => {
+      const ethResults = [
+        '0x000000000000000000000000a78e0e7c9b1b36f7e25c5ccdfdba005ec37eadf4',
+        '0x00000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000000000000000000000000000000000001000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000596f357c6aa34e3d9c3d971121a8616f071b07d2f34fce1da803f7c7ce711e73a28520066f697066733a2f2f516d64745753567a5162555a383269695a4d59684d5657414766446b7732665a52486a634a526e424875617a543800000000000000',
```
CallumGrindle, line 44: Is this second element needed? The only time this array is used is to return the first element in the `mockProvider` so why do we need this?

Esanim: I am using it kinda as `mockResolvedValueOnce` from jest.  The provider is called two times: once for support interface, and the second time for the actual `getData` call. I am calling unshift on the `ethResults` (it's mutable).
Maybe the first mock response (from supportsInterface) should be 
`0x0000000000000000000000000000000000000000000000000000000000000001` instead of 
some random `0x000000000000000000000000a78e0e7c9b1b36f7e25c5ccdfdba005ec37eadf4`.

CallumGrindle: Ah okay nice. makes sense then

CallumGrindle (approved):

frozeman (approved):

PR #130, opened 2022-04-06
Fix npm pack and publish + bump to 0.11.1

Hugoo:
The change in TypeScript structure:

https://github.com/ERC725Alliance/erc725.js/pull/121/files#diff-b55cdbef4907b7045f32cc5360d48d262cca5f94062e353089f189f4460039e0R32

Has changed the structure of the `build/` folder which now has a `schemas/` and `src/` folders.

The `src/` folder was ignored in .npmignore which was the source of a non working release (v0.11.0) because the `build/**/src/` folders were missing.

This also adds a import schemas validation feature which protects the app from crashing when being used with non valid schemas.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,14 @@
 
 All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
+### [0.11.1](https://github.com/ERC725Alliance/erc725.js/compare/v0.11.0...v0.11.1) (2022-04-06)
+
+This version fix the npm pack error.
+
+### Bug Fixes
+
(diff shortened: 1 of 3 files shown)


src/index.ts
```diff
@@ -100,14 +101,36 @@ export class ERC725<Schema extends GenericSchema> {
...
+      schemas: this.validateSchemas(schemas),
       address,
       provider: this.initializeProvider(provider),
       ...defaultConfig,
       ...config,
     };
   }
 
+  /**
+   * To prevent weird behovior from the lib, we must make sure all the schemas are correct before loading them.
```
CallumGrindle, line 113: typo `behaviour`

CallumGrindle (approved):

PR #132, opened 2022-04-26
docs grammar/styling overhaul

fhildeb:
Split off from [docs PR166](https://github.com/lukso-network/docs/pull/166)

Includes:
- Overall grammar work
- applying unified styling
- adding tables for parameters/return values as we do in standard/contract sections
- fixed missing nav position

Explanation of using array[&nbsp;] instead of array[ ]:
Arrays are declared with spaces between brackets to improve clarity, but if you have more than one possible type or a small device, the second bracket gets split up with a line break. That's why we need to use the space symbol to always keep it as one object.

Already applied suggestions from Callum from PR166

### DIFF
--- a/docs/classes/ERC725.md
+++ b/docs/classes/ERC725.md
@@ -1,3 +1,7 @@
+---
+sidebar_position: 1
+---
+
 # ERC725
 
 ## decodeData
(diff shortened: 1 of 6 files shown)


docs/classes/ERC725.md
```diff
@@ -6,27 +10,29 @@
 erc725.decodeData(data);
 ```
 
-In case you are reading the key-value store from an ERC725 smart-contract key-value store
-without `@erc725/erc725.js` you can use `decodeData` to do the decoding for you.
+If you are reading the key-value store from an ERC725 smart-contract key-value store without the `@erc725/erc725.js` library you can use the `decodeData` function to do the decoding for you.
```
CallumGrindle, line 13: I know you didn't edit the language here but I'm confused by this line. How could you use `decodeData` function without using the @erc725/erc725.js library? 

I think maybe it should just be:
`If you are reading the key-value store from an ERC725 smart-contract you can use the `decodeData` function to do the decoding for you.`

docs/classes/ERC725.md
```diff
@@ -6,27 +10,29 @@
...
 
 :::tip
-It is more convenient to use [`fetchData`](ERC725.md#fetchdata).
-It does the `decoding` and `fetching` of external references for you automatically.
+If you want total convenience, it is recommended to use the [`fetchData`](01-ERC725.md#fetchdata) function, which automatically `decodes` and `fetches` external references.
```
CallumGrindle, line 16:
```suggestion
If you want total convenience, it is recommended to use the [`fetchData`](ERC725.md#fetchdata) function, which automatically `decodes` and `fetches` external references.
```

docs/classes/ERC725.md
```diff
@@ -85,26 +93,28 @@ const decodedDataManyKeys = erc725.decodeData({
...
-`Object`
+| Name          | Type   | Description                                                                                     |
+| :------------ | :----- | :---------------------------------------------------------------------------------------------- |
+| `encodedData` | Object | an object with the same keys as the object passed in as a parameter containing the encoded data |
 
-An object with the same keys as the object that was passed in as a parameter containing the encoded data, ready to be stored on the blockchain.
+After the `data` is encoded, the object is ready to be stored in smart contracts.
 
-#### Example
+#### Single-Key Example V1
```
CallumGrindle, line 115: I'd probably remove the `v1` here or just go with `Single-Key Example 1`. It could get confusing as it could be interpreted as a different version of the library

docs/classes/ERC725.md
```diff
@@ -640,33 +690,52 @@ ERC725.encodeKeyName(
...
-2. `signature` - `string`: The raw RLP encoded signature (can be generated with [`web3.eth.accounts.sign()`](https://web3js.readthedocs.io/en/v1.2.11/web3-eth-accounts.html#sign)).
+| Name            | Type   | Description                                           |
+| :-------------- | :----- | :---------------------------------------------------- |
+| `messageOrHash` | string | value of an message or hash that needs to be verified |
+| `signature`     | string | The raw RLP encoded signature                         |
+
+:::info
+
+- The hash must be 66 chars long with the `0x` prefix, otherwise it will be interpreded as message.
+- The message will be: enveloped as `"\x19Ethereum Signed Message:\n" + message.length + message` <br/> and hashed using `keccak256` function.
```
CallumGrindle, line 705: why the line break here?

CallumGrindle (changes requested):
Cool. I left a few suggestions. There is one wrong link which will need to be corrected

CallumGrindle (approved):

docs/classes/ERC725.md
fhildeb: The code format splits up between 2 lines and does not look convenient. I wanted to state that there are two things going on: one line signing, one line hashing

fhildeb:
There actually was a second broken link

Issue #135, opened 2022-05-05
`encodeKeyName` and `decodePermissions`, `encodePermissions` should be available from both class and instance


Hugoo:
Both should work:

```js
ERC725.encodeKeyName(keyName);
```

```js
const erc725 = new ERC725(...);
erc725.encodeKeyName(keyName);
```

same for `ERC725.decodePermissions(permission);`

PR #146, opened 2022-05-16
improvement: add instance versions of static methods

CallumGrindle:
Improvement

Adds instance versions of `encodeKeyName` and `decodePermissions`, `encodePermissions`. Instance versions are available on an instance of ERC725 and call the static version directly.

Closes https://github.com/ERC725Alliance/erc725.js/issues/135

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -52,6 +52,7 @@ import {
   SUPPORTED_HASH_FUNCTION_STRINGS,
 } from './lib/constants';
 import { URLDataWithHash, KeyValuePair } from './types';
+import { Permissions } from './types/Method';
 
 export {
   ERC725JSONSchema,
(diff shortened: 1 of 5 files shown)


src/index.ts
```diff
@@ -577,6 +577,28 @@ export class ERC725<Schema extends GenericSchema> {
...
+    CHANGEOWNER?: boolean;
+    CHANGEPERMISSIONS?: boolean;
+    ADDPERMISSIONS?: boolean;
+    SETDATA?: boolean;
+    CALL?: boolean;
+    STATICCALL?: boolean;
+    DELEGATECALL?: boolean;
+    DEPLOY?: boolean;
+    TRANSFERVALUE?: boolean;
+    SIGN?: boolean;
```
Hugoo, line 597: maybe you can create an interface for this and use the same interface for the static and non static method?

Hugoo (approved):
can you also update the docs to add an example for both the instance and non instance verison?

PR #153, opened 2022-05-20
improvement: add ipfs url conversion

CallumGrindle:
Formats IPFS gateway on instantiation. 

Will add `/ipfs/` path is missing from `ipfsGateway` config option


src/index.ts
```diff
@@ -695,5 +697,3 @@ export class ERC725<Schema extends GenericSchema> {
     return ERC725.encodeKeyName(keyName);
   }
 }
-
-export default ERC725;
```
Hugoo, line 699: sure?

src/index.test.ts
```diff
@@ -353,42 +353,51 @@ describe('Running @erc725/erc725.js tests...', () => {
...
-              name: 'TestJSONURL',
-              key: '0xd154e1e44d32870ff5ade9e8726fd06d0ed6c996f5946dabfdfd46aa6dd2ea99',
-              keyType: 'Singleton',
-              valueContent: 'JSONURL',
-              valueType: 'bytes',
+              returnData: allRawData.filter(
+                (rawData) =>
+                  rawData.key ===
+                  '0xd154e1e44d32870ff5ade9e8726fd06d0ed6c996f5946dabfdfd46aa6dd2ea99',
+              ),
```
Hugoo, line 370: You are already testing all these things in the utils.test.ts:
https://github.com/ERC725Alliance/erc725.js/pull/153/files#diff-c61ef6d0afb4fe242a243493f3db1bedfb8b230688347d7c396d5c1854b11a71R570

Here we only need one custom IPFS format and check if it has been called with the custom format.
If it works with one custom format, no need to test with 4 custom formats.

Checking that the endpoint has been correctly transformed is done in the utils.test already

Hugoo (changes requested):

PR #157, opened 2022-05-23
Dynamic keys should be constructed on the fly

### LINKED ISSUE #137: Dynamic keys should be constructed on the fly with `getData` and `fetchData`
Dynamic keys (any kind, and future kind as well...)
Should be dynamically constructed when passed into `getData`, `fetchData` or `encode/decode` functions, as follows:

```js
LSP3IssuedAssetsMap: {
      dynamicKeyParts: '0xa3e6F38477D45727F6e6f853Cdb479b0D60c0aC9', // or also ['0xa3e6F38477D45727F6e6f853Cdb479b0D60c0aC9']
      value: '0x6e6f853Cdb479b0D60c0aC9'
}
 ```
 
Example for map key with grouping map with grouping
```js
LSP3IssuedAssetsMapWithGrouping: {
      dynamicKeyParts: ['helloWorld', '0xa3e6F38477D45727F6e6f853Cdb479b0D60c0aC9'], // first group word, second group word
      value: '0xb479b0D60c0aC9'
}
```

if value starts with `0x` do not hash, otherwise hash with `keccak256` before constructing the key.

Also constructed keys need to have `0x` removed in the middle.

Hugoo:
Closes #137

### DIFF
--- a/docs/classes/ERC725.md
+++ b/docs/classes/ERC725.md
@@ -619,9 +619,9 @@ Returns the contract owner and is not directly related to ERC725 specifications.
 
 #### Parameters
 
-| Name                 | Type   | Description                               |
-| :------------------- | :----- | :---------------------------------------- |
-| `address` (optional) | string | The contract or EOA address of the owner. |
+| Name                 | Type   | Description                                      |
+| :------------------- | :----- | :----------------------------------------------- |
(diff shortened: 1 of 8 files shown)


Hugoo:
Moved to #158

PR #174, opened 2022-06-07
Refactor API

### LINKED ISSUE #137: Dynamic keys should be constructed on the fly with `getData` and `fetchData`

Hugoo:
Refactor and unify most of the function so they get a similar input (which supports dynamicKeys) and similar output (which is now an array of object rather than an object so it supports returning dynamic keys).

## `encodeData`

```js
erc725.encodeData([
  {
    keyName: 'LSP3Profile',
    value: {
      json: profileJson,
      url: 'ipfs://QmQTqheBLZFnQUxu5RDs8tA9JtkxfZqMBcmGd9sukXxwRm',
    },
  },
]);
```

image

## `decodeData`

```js
const decodedDataOneKey = erc725.decodeData(
  [
    {
      keyName: 'MyKeyName:<bytes32>:<bool>',
      dynamicKeyParts: [
        '0xaaaabbbbccccddddeeeeffff111122223333444455556666777788889999aaaa',
        'true',
      ],
      value: '0x6f357c6a820464ddfac1bec070cc14a8daf04129871d458f2ca94368aae8391311af6361696670733a2f2f516d597231564a4c776572673670456f73636468564775676f3339706136727963455a4c6a7452504466573834554178',
    },
  ],
```

image

## `getData`

image

image

## `fetchData`

image

image

Closes #137


CallumGrindle (approved):
Looks good

src/lib/constants.ts
```diff
@@ -6,7 +6,8 @@ import { MethodData, Encoding, Method } from '../types/Method';
 // https://github.com/ERC725Alliance/ERC725/blob/develop/docs/ERC-725.md#specification
 export const INTERFACE_IDS = {
   ERC725Y_LEGACY: '0x2bd57b73',
-  ERC725Y: '0x5a988c0f',
+  ERC725Y_v200: '0x5a988c0f', // introduced in v2.0.0
+  ERC725Y: '0x714df77c', // introduced in v6.0.0
```
CallumGrindle, line 10: Is this the smart contract version? It is v0.6.0 if so

src/providers/ethereumProviderWrapper.ts
```diff
@@ -67,6 +68,17 @@ export class EthereumProviderWrapper {
...
+    const isErc725Yv200 = await this.supportsInterface(
+      address,
+      INTERFACE_IDS.ERC725Y_v200,
+    );
+
+    if (isErc725Yv200) {
+      return ERC725_VERSION.ERC725;
+    }
+
+    // v2.0.0 and v6.0.0 have the same function signatures for getData, only versions before v2.0.0 requires a different call
```
CallumGrindle, line 80: I think these versions are wrong. Current smart contract version is v0.6.0

CallumGrindle (approved):
Cool! Left minor comments

src/providers/ethereumProviderWrapper.ts
Hugoo: ah indeed my bad, it is v0.2.0 instead of v2.0.0 - thanks will update

CJ42 (commented):
added review comments

src/lib/constants.ts
```diff
@@ -6,7 +6,8 @@ import { MethodData, Encoding, Method } from '../types/Method';
 // https://github.com/ERC725Alliance/ERC725/blob/develop/docs/ERC-725.md#specification
 export const INTERFACE_IDS = {
   ERC725Y_LEGACY: '0x2bd57b73',
-  ERC725Y: '0x5a988c0f',
+  ERC725Y_v200: '0x5a988c0f', // introduced in v0.2.0
```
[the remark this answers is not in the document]
CJ42: I made a suggested PR here: https://github.com/ERC725Alliance/erc725.js/pull/179

PR #194, opened 2022-06-22
Update JsonRpc eth_call parameters

### LINKED ISSUE #193: Get data fails with l16
When using erc725.js to get data the provider fails to get the keys from the erc725 contract

This only seems to happen when using an l16 provider. Tested using same contract version on l14 and worked okay

Getting the same data from l16 using web3 contract instance works okay

image

CallumGrindle:
Fixes issue where data cannot be fetched from l16 closes #193 

eth_call specs are different from what was implemented:
- Added 'latest' tag
- Removed gasPrice so call does not fail with gas error

Tested and this now works with both l14 and l16

https://github.com/ethereum/go-ethereum/issues/21108
https://www.quicknode.com/docs/ethereum/eth_call

### DIFF
--- a/src/lib/provider-wrapper-utils.ts
+++ b/src/lib/provider-wrapper-utils.ts
@@ -64,10 +64,10 @@ export function constructJSONRPC(
       {
         to: address,
         gas: METHODS[method].gas,
-        gasPrice: METHODS[method].gasPrice,
         value: METHODS[method].value,
         data,
       },
+      'latest',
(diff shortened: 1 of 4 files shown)


CallumGrindle:
Still seems to be issues when using browser extension injected ethereum provider. Metamask works okay

Esanim (approved):

Hugoo (approved):

PR #212, opened 2022-07-15
WIP: Add boolean value content encoding

CallumGrindle:
Adds boolean value content encoding.

Need to fix issue as by default all values are decoded as bytes from the provider. This does not work for boolean values and causes encoding failure errors.

`Error: data out-of-bounds (length=31, offset=32, code=BUFFER_OVERRUN, version=abi/5.0.7)
`


CallumGrindle:
Stale and implemented in https://github.com/ERC725Alliance/erc725.js/pull/266. Closing

PR #214, opened 2022-07-22
feat: expose dynamic parts to public encodeKeyName

### LINKED ISSUE #213: encodeKeyName should accept dynamic keys
### I'm submitting a...

- [x] bug report

### Summary

`encodeKeyName` should work for dynamic keys.
The function should accept a second optional parameter `dynamicKeyParts` which is used in case of a dynamic schema.
Current implementation only accepts 1 parameter

image

CallumGrindle:
Adds support for dynamic parts for `encodeKeyName` function. 

Dynamic Parts variable was not exposed to public function call

Closes #213


Hugoo (approved):
nice catch!

PR #229, opened 2022-09-16
fix: Change the reference to old LSP7 interface id to a random value

b00ste:

fix

### Other information:

Change the reference of old LSP7 interface id  to a random value as it is not relevant for the tests.
- `0xe33f65c3` to `0xdeadbeef`

### DIFF
--- a/src/lib/decodeData.test.ts
+++ b/src/lib/decodeData.test.ts
@@ -183,8 +183,8 @@ describe('tuple', () => {
       {
         valueContent: '(Bytes4,Number)',
         valueType: '(bytes4,bytes8)',
-        encodedValue: '0xe33f65c3000000000000000c',
-        decodedValue: ['0xe33f65c3', '12'],
+        encodedValue: '0x5fcaac27000000000000000c',
+        decodedValue: ['0x5fcaac27', '12'],
       },
     ]; // TODO: add more cases? Address, etc.
 
--- a/src/lib/decodeData.test.ts
+++ b/src/lib/decodeData.test.ts
@@ -183,8 +183,8 @@ describe('tuple', () => {
       {
         valueContent: '(Bytes4,Number)',
         valueType: '(bytes4,bytes8)',
-        encodedValue: '0x5fcaac27000000000000000c',
-        decodedValue: ['0x5fcaac27', '12'],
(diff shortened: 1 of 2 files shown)


Hugoo (commented):
These cahnges are on very low level function tests: `TupleKeyValue`

These functions do not care about the content of the tuple. They simply encode/decode. This is not related to any standard.

It happened that we use the test case with values from our standards. But this is really not related. Therefore, I would suggest to simply use totally random value. So no one in the future thinks these tests are related to any LSP / standard, interface ids.

src/lib/utils.test.ts
```diff
@@ -282,8 +282,8 @@ describe('utils', () => {
       {
         valueContent: '(Bytes4,Number)',
         valueType: '(bytes4,bytes8)',
-        encodedValue: '0xe33f65c30000000000000010',
-        decodedValue: ['0xe33f65c3', 16],
+        encodedValue: '0x5fcaac270000000000000010',
+        decodedValue: ['0x5fcaac27', 16],
```
Hugoo, line 286:
```suggestion
        encodedValue: '0xdeadbeaf0000000000000010',
        decodedValue: ['0xdeadbeaf', 16],
```

src/lib/decodeData.test.ts
```diff
@@ -183,8 +183,8 @@ describe('tuple', () => {
       {
         valueContent: '(Bytes4,Number)',
         valueType: '(bytes4,bytes8)',
-        encodedValue: '0xe33f65c3000000000000000c',
-        decodedValue: ['0xe33f65c3', '12'],
+        encodedValue: '0x5fcaac27000000000000000c',
+        decodedValue: ['0x5fcaac27', '12'],
```
Hugoo, line 187:
```suggestion
        encodedValue: '0xdeadbeaf000000000000000c',
        decodedValue: ['0xdeadbeaf', '12'],
```

Hugoo (commented):
lets make it totally different - instead of putting new values from standards.

b00ste:
> lets make it totally different - instead of putting new values from standards.

Tests are passing locally.

Hugoo (approved):

Issue #205, opened 2022-06-29
Add `.supportsInterface()` method


Hugoo:
```js
ERC725.supportsInterface(interfaceId or interfaceName, options) > bool
```

```js
myErc725.supportsInterface(interfaceId or interfaceName) > bool
```

interfaceName [from our [docs](https://docs.lukso.tech/standards/smart-contracts/interface-ids)]:
```
ERC1271,
ERC725X,
ERC725Y,
LSP0ERC725Account,
LSP1UniversalReceiver,
LSP1UniversalReceiverDelegate
LSP6KeyManager,
LSP7DigitalAsset,
LSP8IdentifiableDigitalAsset,
LSP9Vault,
```

`options` :
```
{
  address: '0x23456789',
  provider: 'https://....'
}
```

This object can also be reused for other functions on the class, such as:

```js
ERC725.getOwner({
address: '0x23456789',
provider: 'https://....'
}) > address
```

 // create new issue

On instance, `address` is optional -> defaults to the address used on the instance.

Returns

## Reference:

- https://docs.lukso.tech/standards/smart-contracts/interface-ids

Hugoo:
Or maybe `.detectLSPs` and it does't return ERC725X etc. ?

PR #230, opened 2022-09-19
Add interface and schema detection for contracts

fhildeb:
This PR introduces interface ID and schema key checkups as standalone modular function within the library that can be called directly on the _ERC725 object_ or an _custom smart contract address as parameter_. For convenience, the detection tool can not only handle interface IDs and schema Keys, but also a variety of names. 

### Functions:
```
ERC725.supportsInterface(interfaceId or interfaceName, options) > bool
myErc725.supportsInterface(interfaceId or interfaceName) > bool

ERC725.supportsSchema(schemaKey or schemaName, options) > bool
myErc725.supportsSchema(interfaceId or schemaName) > bool
```

### Default supported interface names: 
```
ERC1271,
ERC725X,
ERC725Y,
LSP0ERC725Account,
LSP1UniversalReceiver,
LSP1UniversalReceiverDelegate
LSP6KeyManager,
LSP7DigitalAsset,
LSP8IdentifiableDigitalAsset,
LSP9Vault
```

### Default supported schema names: 
```
LSP3UniversalProfile,
LSP4DigitalAssetMetadata,
LSP9Vault,
SupportedStandard:LSP3UniversalProfile,
SupportedStandard:LSP4DigitalAssetMetadata,
SupportedStandard:LSP9Vault
```

When using one of the default schema names, the corresponding schema is also loaded directly from the library.

### References

#205 
#235 

### Pending

- [ ] Write documentation
- [ ] Add value/mockup tests


CallumGrindle:
Please remove the package-lock.json from this PR. There are no changes to package.json so we don't expect any changes to package-lock

src/lib/constants.ts
```diff
@@ -169,3 +169,64 @@ export const LSP6_DEFAULT_PERMISSIONS = {
...
+          type: 'bool',
+        },
+      ],
+      stateMutability: 'view',
+      type: 'function',
+    },
+  ],
+};
+
+export const IPFS_GATEWAY_BASE_URL = 'https://ipfs.io/ipfs/'; // 'https://2eff.lukso.dev/ipfs/' not working;
```
CallumGrindle, line 197: Is there not already a default ipfs gateway value somewhere? Why not use that?

src/lib/constants.ts
```diff
@@ -169,3 +169,64 @@ export const LSP6_DEFAULT_PERMISSIONS = {
...
+
+export enum LSPType {
+  LSP3UniversalProfile = 'LSP3',
```
CallumGrindle, line 200: We should use full LSP names

src/lib/constants.ts
```diff
@@ -169,3 +169,64 @@ export const LSP6_DEFAULT_PERMISSIONS = {
...
+  LSP1UniversalReceiver: '0x6bb56a14',
+  LSP1UniversalReceiverDelegate: '0xa245bbda',
+  LSP6KeyManager: '0xc403d48f',
+  LSP7DigitalAsset: '0x5fcaac27',
+  LSP8IdentifiableDigitalAsset: '0x49399145',
+  LSP9Vault: '0xfd4d5c50',
+  ClaimOwnership: '0xa375e9c6',
+};
+
+export interface LspTypeOption {
```
CallumGrindle, line 229: ideally interfaces would go in some `interfaces` file as they are not really constants

src/lib/constants.ts
```diff
@@ -169,3 +169,64 @@ export const LSP6_DEFAULT_PERMISSIONS = {
...
+  ERC721: '0x80ac58cd',
+  ERC721Metadata: '0x5b5e139f',
+  ERC777: '0xe58e113c',
+  ERC1155: '0xd9b67a26',
+  ERC725X: '0x44c028fe',
+  ERC725Y: '0x714df77c',
+  LSP0ERC725Account: '0xeb6be62e',
+  LSP1UniversalReceiver: '0x6bb56a14',
+  LSP1UniversalReceiverDelegate: '0xa245bbda',
+  LSP6KeyManager: '0xc403d48f',
```
CallumGrindle, line 222: Handling versioning for these worries me a bit, what about having an object for these containing the versions or smth?

src/index.ts
```diff
@@ -673,6 +682,98 @@ export class ERC725 {
   encodeKeyName(keyName: string, dynamicKeyParts?: DynamicKeyParts): string {
     return encodeKeyName(keyName, dynamicKeyParts);
   }
-}
 
+  baseConfig = { ipfsGateway: IPFS_GATEWAY_BASE_URL };
```
CallumGrindle, line 686: Can you not use the ipfs config of the ERC725 class rather than recreate it here?

src/index.ts
```diff
@@ -673,6 +682,98 @@ export class ERC725 {
...
+
+  lspTypeOptions: Record<Exclude<LSPType, LSPType.Unknown>, LspTypeOption> = {
```
CallumGrindle, line 688: What is this object? It just contains the interface id which we already have in a constant and the SupportedStandard ABI which should be the same for each standard anyway? I think we can simplify this and do without this object

src/index.ts
```diff
@@ -673,6 +682,98 @@ export class ERC725 {
...
+      return results[0];
+    } catch (error) {
+      return null;
+    }
+  }
+
+  async detectLSPInterface(
+    contractAddress: string,
+    lspType: LSPType,
+    web3Provider: Web3,
```
CallumGrindle, line 734: This means we cant use an ethereum provider. You can use the provider that is set on the ERC725 class

src/index.ts
```diff
@@ -673,6 +682,98 @@ export class ERC725 {
...
+    }
+
+    // EIP-165 detection
+    const contract = new web3Provider.eth.Contract(
+      COMMON_ABIS.supportsInterface as AbiItem[],
+      contractAddress,
+    );
+
+    // Check if the contract implements the LSP interface ID
+    let doesSupportInterface: boolean;
```
CallumGrindle, line 747: what about `isInterfaceSupported`?

src/index.ts
```diff
@@ -673,6 +682,98 @@ export class ERC725 {
...
+    try {
+      doesSupportInterface = await contract.methods.supportsInterface(
+        this.lspTypeOptions[lspType],
+      ).call;
```
CallumGrindle, line 751: Does this not need to be called? It looks like this sets the `doesSupportInterface` to the `call` function. Might be a web3 thing. Can you check?

CallumGrindle (changes requested):
Nice function. Though I don't think we should be assuming these standards necessarily coexist and e.g. that an LSP0ERC725Account necessarily supports LSP3. This will lead to confusing outcomes as I could have a valid anon ERC725Account and this function would tell me it is not supported as it does not implement LSP3. 

We should treat each standard separately as they are modular and each standard can be used independently and in different combinations.

e.g. an LSP7 contract does not need to implement LSP4 and LSP4 can be used on a contract which is not LSP7 or LSP8. 

I think decoupling these standards will simplify this function for you as you will not need to check these metadata standards unless LSP3,LSP4,LSP9 is specifically being asked for

src/lib/constants.ts
fhildeb: for some reason, the cloudflare IPFS does not work sometimes for assets on l16. had this issue with Jean several times, so i just thought I go with how we handle it in the wallet.

src/index.ts
fhildeb: We wanted to have one object that can hold interfaceId but also Schema, so the function works pretty generic, and we just check if its a schema or interface standard- and then we basically map everything thats needed onto the LSPs.

src/index.ts
fhildeb: we can leave it out, initially i planned on having a boolean but in the latest version due to modularity, we just return plain values

src/index.ts
fhildeb: it will just return the value of the supportsInterface function

fhildeb:
Split into multiple PRs:

#243 --> merged
#244 --> open discussion

Closed.

PR #236, opened 2022-09-27
feat: make library compatible with RPC urls

skimaharvey:


Hugoo:
Let's not forget to update the docs to reference this new feature.
https://app.clickup.com/t/30jb7we

Hugoo (approved):

PR #239, opened 2022-09-30
refactor: use web3-providers-http instead of web3

skimaharvey:


src/index.ts
```diff
@@ -61,6 +60,8 @@ import { getDataFromExternalSources } from './lib/getDataFromExternalSources';
 import { DynamicKeyParts } from './types/dynamicKeys';
 import { getData } from './lib/getData';
 
+const HttpProvider = require('web3-providers-http');
```
CallumGrindle, line 63: Can we use import for this?

skimaharvey: I tried but was having weird issues
Screenshot 2022-09-29 at 22 52 12
Screenshot 2022-09-29 at 23 00 33
Screenshot 2022-09-29 at 23 02 56

Hugoo (approved):

Issue #192, opened 2022-06-22
Wrong return type for `fetchData`


CallumGrindle:
When fetching profile await erc725.fetchData("LSP3Profile") expected return type is DecodeDataOutput with value field. But value  is typed as string | string[] | URLDataWithHash while in value  I get profile object

```
LSP3Profile: {
...
}
```

![image](https://user-images.githubusercontent.com/54543428/174992296-8089cdb0-3f78-476b-9871-273679354bee.png)

![image](https://user-images.githubusercontent.com/54543428/174992436-c9a90d83-09fc-4775-ab80-6662759c1d46.png)

dzbo:
@Hugoo @CallumGrindle this is pretty old, any plans on pushing this further?

Hugoo:
Hey @dzbo we will take this one for the current sprint

PR #247, opened 2022-10-27
Fix return type for `fetchData`

magalimorin18:
Fix the type of the object returned by fetchData function.

When fetching profile `await erc725.fetchData("LSP3Profile")` expects that it will return `DecodeDataOutput` with `value` field. But value is typed as `string | string[] | URLDataWithHash` while in `value`  I get profile object
```
{
LSP3Profile: 
{...}
}
```
[Link to open issue](https://github.com/ERC725Alliance/erc725.js/issues/192)

A new type `FetchDataOutput` was created with the following properties :
```
FetchDataOutput {
  value:
    | string
    | string[]
    | { LSP3Profile: Record<string, any> }
    | Record<string, any>;
  name: string;
  key: string;
}
```

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -52,6 +52,7 @@ import {
   DecodeDataInput,
   DecodeDataOutput,
   EncodeDataInput,
+  FetchDataOutput,
 } from './types/decodeData';
 import { GetDataDynamicKey, GetDataInput } from './types/GetData';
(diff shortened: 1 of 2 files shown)


Hugoo:
Closes #192

Hugoo (approved):

PR #255, opened 2022-11-08
docs: update docs

Hugoo:
To sync with: https://github.com/lukso-network/docs/pull/434

### DIFF
--- a/docs/providers.md
+++ b/docs/providers.md
@@ -20,15 +20,15 @@ const RPC_URL = 'https://rpc.l16.lukso.network';
 const erc725 = new ERC725([], '0x...', RPC_URL);
 ```
 
-## Ethereum (injected provider from extension)
+## Ethereum (MetaMask)
 
-```javascript
-import ERC725 from '@erc725/erc725.js';
+:::caution Warning
 
-const ethereumProvider = window.ethereum;
+Ethereum providers are being deprecated. Please provide an RPC URL instead.
 
-const erc725 = new ERC725([], '0x...', ethereumProvider);
-```
+:::
+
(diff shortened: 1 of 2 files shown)


docs/providers.md
```diff
@@ -20,15 +20,15 @@ const RPC_URL = 'https://rpc.l16.lukso.network';
...
 
-## Ethereum (injected provider from extension)
+## Ethereum (MetaMask)
 
-```javascript
-import ERC725 from '@erc725/erc725.js';
+:::caution Warning
 
-const ethereumProvider = window.ethereum;
+Ethereum providers are being deprecated. Please provide an RPC URL instead.
```
CallumGrindle, line 27: I don't understand why we are saying ethereum providers are being deprecated. Its a valid way to use the library without maintaining multiple RPC urls

Hugoo: It is easier for us to request an RPC url from the user and start our own ethereum provider rather then having to support multiple ethereum provider. Bc we don't know the versions, etc.

### With URL:

```
new erc725('URL')
    -> in the lib: new Web3(URL) // -> we fully control this
```

### With ethereum provider:

```
new erc725(provider)
      -> is this provider supported?
```

CallumGrindle: Cool thanks, I agree this makes sense for library providers (`ethers.provider.JSONRprProvider` or `Web3.providers.HttpProvider`) but passing in an injected `window.ethereum` object still seems to me to be a valid way to use the lib and one we should continue to support. wdyt?

e.g:
```js
const erc725 = new ERC725([], '0x...', window.ethereum);
```

CallumGrindle (approved):

PR #256, opened 2022-11-28
fix: Encode key name should parse any number (hex or decimal) for uint type

lykhonis:
bug fix

uint type takes only decimals

uint type can take hex and decimals

### Other information:
N/A


CallumGrindle (approved):

CallumGrindle:
Nice, thanks for sending!

Issue #252, opened 2022-10-28
Add `tuple` and `CompactBytesArray` decoding


frozeman:
Should be able to decode `valueType` and `valueContent` of **tuples**:

`(bytes4,address,bytes4)` results in -> `["0x11223344", "0x1122334411223344112233441122334411223344", "0x11223344"]`

**CompactBytesArray:**
first byte length of bytes l, followed by bytes, follo3ed by length of next bytes.
`valueContent  = "CompactBytesArray"`
"0x**04**11223344**14**1122334411223344112233441122334411223344**04**11223344" in     
-> `["0x11223344", "0x1122334411223344112233441122334411223344", "0x11223344"]`

Should also work:
`(Bytes4,Address,Bytes4)[]`  // ABI encoded array of a tuple
`(Bytes4,Address,Bytes4)[CompactBytesArray]` // CompactBytesArray encoded tuple (cheaper)

Hugoo:
Closed in https://github.com/ERC725Alliance/erc725.js/pull/261

PR #261, opened 2023-01-19
feat: add  encoding / decoding for `bytes[CompactBytesArray]`

CJ42:
Co-authored-by: b00ste.lyx <b00ste@users.noreply.github.com>

⭐ feature

There is currently no support for encoding [`CompactBytesArray`](https://github.com/lukso-network/LIPs/blob/main/LSPs/LSP-2-ERC725YJSONSchema.md#bytescompactbytesarray) defined in LSP2.

Related to issue #252 

Add basic support for `bytes[CompactBytesArray]`


src/lib/encoder.ts
```diff
@@ -132,6 +132,54 @@ const valueTypeEncodingMap = {
...
+    let hexNumber = padLeft(numberToHex(numberOfBytes), 4);
+
+    compactBytesArray += hexNumber.substring(2) + values[i].substring(2);
+  }
+
+  return compactBytesArray;
+};
+
+const decodeCompactBytesArray = (compactBytesArray: string): string[] => {
+  if (!isHex(compactBytesArray))
```
skimaharvey: does `isHex` returns true on empty string? because at the SC level we consider 0x to be a valid CBA and we may have keys with empty CBA ? could not see a test for that)

skimaharvey (approved):
LGTM. May be review the comment about empty bytes arrays

src/lib/encoder.ts
CJ42: @skimaharvey added test for this case scenario

src/lib/encoder.ts
```diff
@@ -80,6 +81,55 @@ const decodeDataSourceWithHash = (value: string): URLDataWithHash => {
   return { hashFunction: hashFunction.name, hash: dataHash, url: dataSource };
 };
 
+const encodeCompactBytesArray = (values: string[]): string => {
+  let compactBytesArray = '0x';
+
+  for (let i = 0; i < values.length; i++) {
+    if (values[i] === '')
```
CallumGrindle, line 88: Doesn't the next `isHex` cover this anyway?

CallumGrindle (commented):
I lack some bytes know how here but lgtm. Please update the docs too :)

src/lib/encoder.ts
b00ste: `isHex('')` actually returns `true`

magalimorin18 (commented):
Looks good to me too,
Very clear and easy to read code Thank you!

src/lib/encoder.test.ts
```diff
@@ -117,6 +117,32 @@ describe('encoder', () => {
...
+          // 500 bytes long
+          '0xbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeef',
```
Hugoo, line 136:
```js
// 512 bytes long value
          `0x${'cafe'.repeat(256)}`,
          // 500 bytes long
          `0x${'beef'.repeat(250)}`,
```

would make it easier to read and faster to see it is indeed 512 and 500 bytes long

src/lib/encoder.test.ts
```diff
@@ -117,6 +117,32 @@ describe('encoder', () => {
...
...fecafecafecafecafecafecafecafecafecafecafecafecafecafecafecafecafecafecafecafecafecafecafecafecafecafecafecafecafecafecafecafecafecafecafecafecafecafecafecafecafecafecafecafecafecafecafecafecafe01f4beefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeefbeef',
```
Hugoo, line 139: same here pls use string concat with repeat. It will make this much easier to read and understand. And also the file will look less ugly.

src/lib/encoder.ts
```diff
@@ -80,6 +81,55 @@ const decodeDataSourceWithHash = (value: string): URLDataWithHash => {
...
+      throw new Error(
+        `Couldn't encode, value at index ${i} is '', expected '0x'`,
```
Hugoo, line 90: is "expected 0x" really true?

Do you mean "Expacted an hexadecimal string with 0x prefix instead?

src/lib/encoder.ts
```diff
@@ -80,6 +81,55 @@ const decodeDataSourceWithHash = (value: string): URLDataWithHash => {
...
+      );
+
+    const numberOfBytes = (values[i].length - 2) / 2;
+    const hexNumber = padLeft(numberToHex(numberOfBytes), 4);
+
+    compactBytesArray += stripHexPrefix(hexNumber) + stripHexPrefix(values[i]);
+  }
+
+  return compactBytesArray;
+};
```
Hugoo, line 108: I would suggest to make this more "javascript like" by using filter and reduce instead of a for.
The code bellow should work:

```suggestion
const encodeCompactBytesArray = (values: string[]): string => {
  const compactBytesArray = values
    .filter((value, index) => {
      if (value === '') {
        throw new Error(
          `Couldn't encode, value at index ${index} is '', expected '0x'`,
        );
      }

      if (!isHex(value)) {
        throw new Error(`Couldn't encode, value at index ${index} is not hex`);
      }

      if (value.length > 65_535 * 2 + 2) {
        throw new Error(
          `Couldn't encode bytes[CompactBytesArray], value at index ${index} exceeds 65_535 bytes`,
        );
      }

      return true;
    })
    .reduce((acc, value) => {
      const numberOfBytes = (value.length - 2) / 2;
      const hexNumber = padLeft(numberToHex(numberOfBytes), 4);

      return acc + stripHexPrefix(hexNumber) + stripHexPrefix(value);
    }, '0x');

  return compactBytesArray;
};
```

src/lib/encoder.ts
```diff
@@ -80,6 +81,55 @@ const decodeDataSourceWithHash = (value: string): URLDataWithHash => {
...
+
+    if (values[i].length > 65_535 * 2 + 2)
+      throw new Error(
+        `Couldn't encode bytes[CompactBytesArray], value at index ${i} exceeds 65_535 bytes`,
+      );
+
+    const numberOfBytes = (values[i].length - 2) / 2;
+    const hexNumber = padLeft(numberToHex(numberOfBytes), 4);
+
+    compactBytesArray += stripHexPrefix(hexNumber) + stripHexPrefix(values[i]);
```
Hugoo, line 104: Here you do `-2` because you want to remove the length of `0x` from the string?
Then you use stripHexPrefix. So you assume all the hex string will start with `0x`. But you are using `isHex` (and not `isHexStrict` https://web3js.readthedocs.io/en/v1.2.11/web3-utils.html#ishexstrict

So I reco:

- Either you use isHexString
- or you make your code work for both 0x and non 0x prefixed hex string.

I recommend the second option.

src/lib/encoder.test.ts
```diff
@@ -117,6 +117,32 @@ describe('encoder', () => {
         encodedValue:
           '0x000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000020000000000000000000000000000000000000000000000000000000000000040000000000000000000000000000000000000000000000000000000000000008000000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000000000000000000000000000000001337000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000054ef',
       },
+      {
+        valueType: 'bytes[CompactBytesArray]',
+        decodedValue: [
+          '0xaabb',
+          '0xcafecafecafecafecafecafecafecafecafecafecafecafecafecafecafecafe',
```
Hugoo, line 124: if you remove `0x` from this string, the test will fail, but I think they should not. Cf my comment bellow on isHexStrict

src/lib/encoder.ts
```diff
@@ -80,6 +81,55 @@ const decodeDataSourceWithHash = (value: string): URLDataWithHash => {
...
+  }
+
+  return compactBytesArray;
+};
+
+const decodeCompactBytesArray = (compactBytesArray: string): string[] => {
+  if (!isHex(compactBytesArray))
+    throw new Error("Couldn't decode, value is not hex");
+
+  let pointer = 2;
```
Hugoo, line 114: I think this might also fail by using non `0x` prefixed hex strings as it will pass the `isHex` and then you will remove 2 chars

src/lib/encoder.ts
b00ste: Yeah, if someone passes '', it is not valid hex value and `isHex('')` returns `true`. When encoding a '' value it has unexpected behaviour and if you want to encode an empty bytes you need to pass it as '0x' and not ''

src/lib/encoder.ts
CJ42: nice will commit that! 👍

src/lib/encoder.ts
CJ42: @Hugoo I am also in favour of option 2.
The problem is that in the test suite.

Screenshot 2023-01-24 at 08 44 02

it fails because when it decodes, it decodes '' (empty string) as '0x`. And the test states that when decoding it should return the expected values mentioned under the `decodedValue`. So '' (empty string) never match and is always expected to be decoded as `0x`

image

CJ42: conclusion: we keep `isHex()`

Hugoo (approved):

Hugoo:
This is weird, like why would the test coverage change on these files? 🤔 
image

PR #262, opened 2023-01-20
feat: add encoding/decoding for other types of compactBytesArray

b00ste:
Co-authored-by: Jean Cvllr <CJ42@users.noreply.github.com>

Feat, adding the new types for encoding/decoding CompactBytesArray

Current behaviour only allows encoding/decoding for bytes[CompactBytesArray]

The following new types are added:
- bytesN[CompactBytesArray] (e.g. bytes1[CompactBytesArray] to bytes32[CompactBytesArray])
- uintN[CompactBytesArray] (e.g. uint8[CompactBytesArray] to uint256[CompactBytesArray])
- string[CompactBytesArray]


src/lib/encoder.test.ts
```diff
@@ -201,6 +221,53 @@ describe('encoder', () => {
       });
     });
 
+    describe('when encoding uintN[CompactBytesArray]', () => {
+      it('should thow if trying to encode a value that exceeds its type max bytes length', async () => {
```
magalimorin18, line 225: throw instead of thow

magalimorin18: the maximal lenght of bytes for this type

src/lib/encoder.test.ts
```diff
@@ -201,6 +221,53 @@ describe('encoder', () => {
...
+      });
+
+      it('should thow if trying to decode a value that exceeds its type max bytes length', async () => {
+        expect(() => {
+          decodeValueType(
+            'uint8[CompactBytesArray]',
+            '0x00010100012b00014900020100',
+          );
+        }).to.throw(
+          'Hex uint8 value at index: 3 is using 2 bytes, which exceedes 1',
```
magalimorin18, line 240: exceeds instead of exceedes

src/lib/encoder.test.ts
```diff
@@ -201,6 +221,53 @@ describe('encoder', () => {
...
+    describe('when encoding uintN[CompactBytesArray]', () => {
+      it('should thow if trying to encode a value that exceeds its type max bytes length', async () => {
+        expect(() => {
+          encodeValueType('uint8[CompactBytesArray]', [15, 178, 266]);
+        }).to.throw(
+          'Hex uint8 value at index: 2 is using 1.5 bytes, which exceedes 1',
+        );
+      });
+
+      it('should thow if trying to decode a value that exceeds its type max bytes length', async () => {
```
magalimorin18, line 233: throw

src/lib/encoder.ts
```diff
@@ -143,6 +143,123 @@ const decodeCompactBytesArray = (compactBytesArray: string): string[] => {
...
+const encodeBytesNCompactBytesArray = (
+  values: string[],
+  numberOfBytes: number,
+): string => {
+  for (let i = 0; i < values.length; i++) {
+    if (values[i].length > numberOfBytes * 2 + 2)
+      throw new Error(
+        `Hex bytes${numberOfBytes} value at index: ${i} is using ${
+          (values[i].length - 2) / 2
+        } bytes, which exceedes ${numberOfBytes}`,
```
magalimorin18, line 155: Here it should be exceeds and not exceedes no ?

b00ste: Yep, good catch

src/lib/encoder.ts
```diff
@@ -143,6 +143,123 @@ const decodeCompactBytesArray = (compactBytesArray: string): string[] => {
...
+        `Hex bytes${numberOfBytes} value at index: ${i} is using ${
+          (bytesValues[i].length - 2) / 2
+        } bytes, which exceedes ${numberOfBytes}`,
+      );
+  }
+
+  return bytesValues;
+};
+
+const returnCompactBytesArrayBytesNTypes = () => {
```
magalimorin18, line 180: The name of the function is not very clear to what it is doing ?

magalimorin18 (commented):
Very cool thanks :)

src/lib/encoder.ts
b00ste: yep, sorry, it just returns:
{
    'uint8[CompactBytesArray]' = {
        encode: (value: number[]) => encodeUintNCompactBytesArray(value, 1),
        decode: (value: string) => decodeUintNCompactBytesArray(value, 1),
},
{
    'uint16[CompactBytesArray]' = {
        encode: (value: number[]) => encodeUintNCompactBytesArray(value, 2),
        decode: (value: string) => decodeUintNCompactBytesArray(value, 2),
},
[...]
{
    'uint256[CompactBytesArray]' = {
        encode: (value: number[]) => encodeUintNCompactBytesArray(value, 32),
        decode: (value: string) => decodeUintNCompactBytesArray(value, 32),
}

src/lib/encoder.ts
```diff
@@ -143,6 +143,123 @@ const decodeCompactBytesArray = (compactBytesArray: string): string[] => {
   return encodedValues;
 };
 
+const encodeBytesNCompactBytesArray = (
+  values: string[],
+  numberOfBytes: number,
+): string => {
+  for (let i = 0; i < values.length; i++) {
```
CallumGrindle, line 150: cant you do `values.forEach()`?

b00ste: good catch, I made all the functions more JS like

CJ42 (commented):
added review comments

JeneaVranceanu:
Very curious about what the encoding of the compact bytes array of arrays will look like.
The cause for that is the following:
- using ABI encoder for dynamic bytes will produce a value that contains `<size offset> + <size bytes> + <bytes value> [+ <bytes value offset>]`. The last part is not required if the `<bytes value>` is a multiple of 32, otherwise the offset just fills the gap;
- using ABI encoder for tuples will add padding to subtypes where required as well. So a tuple of `(boolean,bytes4)` will be 64 bytes long with and constructed as follows: `<31 zero bytes offset> + <1 byte boolean value> + <4 bytes value> + <28 zero bytes offset>`.

All these offsets are omitted when we encode a compact byte array of bytes or tuples.
But what will be a compact bytes array of arrays?

Executing `web3.eth.abi.encodeParameter('bytes[]',['0x43'])`
Gives the following:
```
'0x00000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000000000000000000000000000000000001000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000014300000000000000000000000000000000000000000000000000000000000000'
```

Where 
```
0000000000000000000000000000000000000000000000000000000000000020 - offset which tells how many bytes to skip before reading the array size
0000000000000000000000000000000000000000000000000000000000000001 - array size
0000000000000000000000000000000000000000000000000000000000000020 - offset which tells how many bytes to skip before reading the first element size
0000000000000000000000000000000000000000000000000000000000000001 - first element size
4300000000000000000000000000000000000000000000000000000000000000 - first element
```

Does something happen to the offsets in this case? Maybe I got a bit carried away but I'm certainly not sure 🙂

JeneaVranceanu:
Additional question: if the compact bytes array has 2 or more levels of nested types do we compact only the first layer or all of them?

Example: `(address[],(boolean,bytes4))[CompactBytesArray]`. It doesn't really make sense but still possible.

Do we encode `...,(boolean,bytes4)...` tuple the default way or do we reduce it as well?

My understanding is that we encode it the default way. The same applies in my opinion to the question above about arrays in CompactBytesArray. 

Encoding example:
```
web3.eth.abi.encodeParameter('address[]', ['0x98d2fF3907A4a9dEb10B1F2F79EBF078984501BF'])
=>
0x0000000000000000000000000000000000000000000000000000000000000020
0000000000000000000000000000000000000000000000000000000000000001
00000000000000000000000098d2ff3907a4a9deb10b1f2f79ebf078984501bf

web3.eth.abi.encodeParameter('(bool,bytes4)', [true, '0xffbbffbb'])
=>
0x0000000000000000000000000000000000000000000000000000000000000001
ffbbffbb00000000000000000000000000000000000000000000000000000000

web3.eth.abi.encodeParameter('(address[],(bool,bytes4))[]', [[['0x98d2fF3907A4a9dEb10B1F2F79EBF078984501BF'], [true, '0xffbbffbb']]])
=>
// And instead of encoding these values as default array (with total count of bytes 256):
0x0000000000000000000000000000000000000000000000000000000000000020
0000000000000000000000000000000000000000000000000000000000000001
0000000000000000000000000000000000000000000000000000000000000020
0000000000000000000000000000000000000000000000000000000000000060
0000000000000000000000000000000000000000000000000000000000000001
ffbbffbb00000000000000000000000000000000000000000000000000000000
0000000000000000000000000000000000000000000000000000000000000001
00000000000000000000000098d2ff3907a4a9deb10b1f2f79ebf078984501bf

// We will encode these values as CompactBytesArray:

00C0 // 192 bytes first value length with total count of bytes = 196
0000000000000000000000000000000000000000000000000000000000000020
0000000000000000000000000000000000000000000000000000000000000060
0000000000000000000000000000000000000000000000000000000000000001
ffbbffbb00000000000000000000000000000000000000000000000000000000
0000000000000000000000000000000000000000000000000000000000000001
00000000000000000000000098d2ff3907a4a9deb10b1f2f79ebf078984501bf
```

CJ42 (approved):

src/lib/encoder.ts
```diff
@@ -143,6 +143,109 @@ const decodeCompactBytesArray = (compactBytesArray: string): string[] => {
...
+  return encodeCompactBytesArray(values);
+};
+
+const decodeBytesNCompactBytesArray = (
+  compactBytesArray: string,
+  numberOfBytes: number,
+): string[] => {
+  const bytesValues = decodeCompactBytesArray(compactBytesArray);
+  bytesValues.forEach((bytesValue, index) => {
+    if (bytesValue.length > numberOfBytes * 2 + 2)
```
samuel-videau, line 166: Yeah and I see it's not only in this PR, but the maybe we should add a comment explaining why there is a `+2`, which I guess rely for the `0x`

src/lib/encoder.ts
```diff
@@ -143,6 +143,109 @@ const decodeCompactBytesArray = (compactBytesArray: string): string[] => {
...
+      throw new Error(
+        `Hex uint${
+          numberOfBytes * 8
+        } value at index ${index} does not fit in ${numberOfBytes} bytes`,
+      );
+    return hexToNumber(hexValue);
+  });
+};
+
+const returnTypesOfUintNCompactBytesArray = () => {
```
samuel-videau, line 225: What about adding JSDocs to the functions, it's an open source project so I guess it's important to add JSDocs and comments for people to properly understand what's going on

src/lib/encoder.ts
```diff
@@ -143,6 +143,109 @@ const decodeCompactBytesArray = (compactBytesArray: string): string[] => {
...
+        `Hex bytes${numberOfBytes} value at index ${index} does not fit in ${numberOfBytes} bytes`,
+      );
+  });
+
+  return bytesValues;
+};
+
+const returnTypesOfBytesNCompactBytesArray = () => {
+  const types = {};
+  for (let i = 1; i < 33; i++) {
```
samuel-videau, line 177: Can't it go higher than 32? as between each value we have a bytes2 (65535)

samuel-videau (changes requested):

src/lib/encoder.ts
b00ste: I can but then it's not a valid Solidity type, higher than 32 is just bytes.

src/lib/encoder.ts
b00ste: Removed +2, replaced with `stripHexPrefix(..)` which should allow both bytes with 0x and without

src/lib/encoder.ts
```diff
@@ -143,6 +143,109 @@ const decodeCompactBytesArray = (compactBytesArray: string): string[] => {
   return encodedValues;
 };
 
+const encodeBytesNCompactBytesArray = (
+  values: string[],
+  numberOfBytes: number,
+): string => {
+  values.forEach((value, index) => {
+    if (value.length > numberOfBytes * 2 + 2)
```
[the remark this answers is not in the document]
b00ste: Removed +2, replaced with stripHexPrefix(..) which should allow both bytes with 0x and without

src/lib/encoder.ts
b00ste: Yeah, good idea

src/lib/encoder.ts
samuel-videau: Alright I see

src/lib/encoder.ts
Hugoo: @magalimorin18 @b00ste can we resolve this comment?

b00ste: I changed the function name and added JSdocs, I would say yes on my end, wdyt @magalimorin18 ?

src/lib/encoder.ts
```diff
@@ -143,6 +153,149 @@ const decodeCompactBytesArray = (compactBytesArray: string): string[] => {
...
+    };
+  }
+  return types;
+};
+
+/**
+ * Encode any set of strings to string[CompactBytesArray]
+ * @param values An array of any strings
+ * @returns Encoding of `values` to string[CompcatBytesArray]
+ */
```
Hugoo, line 280: there is a typo

```suggestion
/**
 * Encodes any set of strings to string[CompactBytesArray].
 *
 * @param values {string[]}
 * @returns Encoding of `values` to string[CompactBytesArray]
 */
```

src/lib/encoder.ts
```diff
@@ -143,6 +153,149 @@ const decodeCompactBytesArray = (compactBytesArray: string): string[] => {
...
+const encodeStringCompactBytesArray = (values: string[]): string => {
+  const hexValues: string[] = values.map((element) => utf8ToHex(element));
+
+  return encodeCompactBytesArray(hexValues);
+};
+
+/**
+ * Decode a string[CompactBytesArray] to an array of strings
+ * @param compactBytesArray A valid string[CompcatBytesArray]
```
Hugoo, line 289:
```suggestion
 * @param compactBytesArray A valid string[CompactBytesArray]
```

src/lib/encoder.ts
```diff
@@ -143,6 +153,149 @@ const decodeCompactBytesArray = (compactBytesArray: string): string[] => {
...
+        } value at index ${index} does not fit in ${numberOfBytes} bytes`,
+      );
+    return hexNumber;
+  });
+
+  return encodeCompactBytesArray(hexValues);
+};
+
+/**
+ * Decode CompactBytesArray of type uintN
```
Hugoo, line 239:
```suggestion
 * Decodes CompactBytesArray of type uintN.
```

src/lib/encoder.ts
```diff
@@ -143,6 +153,149 @@ const decodeCompactBytesArray = (compactBytesArray: string): string[] => {
...
+    types[`bytes${i}[CompactBytesArray]`] = {
+      encode: (value: string[]) => encodeBytesNCompactBytesArray(value, i),
+      decode: (value: string) => decodeBytesNCompactBytesArray(value, i),
+    };
+  }
+  return types;
+};
+
+/**
+ * Encode uintN to CompactBytesArray
```
Hugoo, line 212:
```suggestion
 * Encodes uintN to CompactBytesArray.
```

Hugoo (approved):

PR #264, opened 2023-01-24
feat: add decoding/encoding support for tuples of CompactBytesArray

b00ste:
Co-authored-by: Jean Cvllr <CJ42@users.noreply.github.com>

This PR is a feature.

Current behaviour allows for encoding/decoding of normal Tuples.

The new behaviour allows for encoding/decoding of (Tuples)[CompactBytesArray]


src/lib/utils.test.ts
```diff
@@ -146,6 +146,84 @@ describe('utils', () => {
         );
       });
     });
+
+    it(`encodes a CompactBytesArray of tuple (bytes4,address,bytes2)`, () => {
```
Hugoo, line 150:
```suggestion
    it('encodes a CompactBytesArray of tuple (bytes4,address,bytes2)', () => {
```

lets use back quote only when needed

src/lib/utils.test.ts
```diff
@@ -146,6 +146,84 @@ describe('utils', () => {
...
+          ],
+        ],
+        encodedValue:
+          '0x001ccafecafe951a5d121531bba8DAFEA492D9c6733ae3d56b7Ed1ADB606001cbeefbeef8a483080f5db1105FE31320faF8Da1492Eadf8Deb79bd264001cf00df00d2fe92a11caf28ab2c527702b14BF2f79F70B32e09F62B6A7',
+      };
+
+      assert.deepStrictEqual(
+        encodeKey(
+          testCase.schema as ERC725JSONSchema,
+          testCase.decodedValue as any,
```
Hugoo, line 222: try to avoid using 'as xxx' you very likely can type these values earlier, when you create them.

src/lib/utils.test.ts
```diff
@@ -146,6 +146,84 @@ describe('utils', () => {
...
+        encodedValue:
+          '0x001acafecafeDAFEA492D9c6733ae3d56b7Ed1ADB60692c98Bc5cafe001abeefbeefFE31320faF8Da1492Eadf8Deb79bd264D7cF2141beef001af00df00dc527702b14BF2f79F70B32e09F62B6A74cADFd80f00d'.toLowerCase(),
+      };
+
+      assert.deepStrictEqual(
+        encodeKey(
+          testCase.schema as ERC725JSONSchema,
+          testCase.decodedValue as any,
+        ),
+        testCase.encodedValue,
```
Hugoo, line 185: i don't think you need to make a "testCase" object, just create 3 const and type them:

```
const schema: ERC725JSONSchema = {// the shema}
const decodedValues = [[...]] // TS should detect it is a string[][] type, if not type it here
...
```

src/lib/decodeData.test.ts
```diff
@@ -279,4 +284,97 @@ describe('tuple', () => {
...
+        ],
+      ],
+      encodedValue:
+        '0x001ccafecafe951a5d121531bba8DAFEA492D9c6733ae3d56b7Ed1ADB606001cbeefbeef8a483080f5db1105FE31320faF8Da1492Eadf8Deb79bd264001cf00df00d2fe92a11caf28ab2c527702b14BF2f79F70B32e09F62B6A7',
+    };
+
+    expect(
+      decodeKey(testCase.schema as ERC725JSONSchema, testCase.encodedValue),
+    ).to.deep.equal(testCase.decodedValue);
+  });
```
Hugoo, line 379: It looks like you are doing the same test 3 times.

make a testCases variable and use forEach - as done elsewhere in the project.

Hugoo (changes requested):
i think you can do it without using as xxx - or with fewer of them.
Don't repeat yourself - put the tests with a forEach.

src/lib/decodeData.test.ts
b00ste: I see what you mean now, I moved the 3 tests inside the above forEach which does both encoding and decoding

magalimorin18 (commented):
Nice

src/lib/utils.test.ts
```diff
@@ -132,6 +132,90 @@ describe('utils', () => {
...
+            '0xFE31320faF8Da1492Eadf8Deb79bd264',
+          ],
+          [
+            '0xf00df00d',
+            '0x2fe92a11caf28ab2',
+            '0xc527702b14BF2f79F70B32e09F62B6A7',
+          ],
+        ],
+        encodedValue:
+          '0x001ccafecafe951a5d121531bba8DAFEA492D9c6733ae3d56b7Ed1ADB606001cbeefbeef8a483080f5db1105FE31320faF8Da1492Eadf8Deb79bd264001cf00df00d2fe92a11caf28ab2c527702b14BF2f79F70B32e09F62B6A7',
```
Hugoo, line 189: this should be lower case - and test should work with the lowercase version of this string

Hugoo (changes requested):

src/lib/utils.ts
```diff
@@ -282,23 +283,44 @@ export function encodeKey(
...
+          '[CompactBytesArray]',
+          '',
+        );
+
+        if (
+          Array.isArray(value[0]) &&
+          valueType[0] === '(' &&
+          valueType[valueType.length - 1] === ')'
+        ) {
+          const encodedTuples = (value as string[][]).map((element) => {
```
Hugoo, line 298: whenever you write `as xxx` you are "forcing" TypeScript to cast a value to a certain type.
Very likely, you are doing something wrong or not the best way.

Adding these "as" makes the TypeScript types less relevant and more fragile.

Here, at this stage, TS thinks this value is:
image

You can see in the screenshot that something is probably off.
TypeScript thinks this variable can be of multiple types, but you force it to be one of them.

The best way to have TS be sure that the type is string[][], is to do the propre __JavaScript__ checks.

src/lib/utils.ts
```diff
@@ -282,23 +283,44 @@ export function encodeKey(
...
     case 'mapping':
+      if (schema.valueType.includes('[CompactBytesArray]')) {
+        const valueType = schema.valueType.replace('[CompactBytesArray]', '');
+        const valueContent = schema.valueContent.replace(
+          '[CompactBytesArray]',
+          '',
+        );
+
+        if (
+          Array.isArray(value[0]) &&
```
Hugoo, line 294: By doing `value[0]` you assume that value is an array, but it might not be, as TS shows:
image

So what happens if value is a `number` ? Will it break when you try to call

By adding a first check on the value itself:
image

You make TS smarter, as you can see:
image


Now the set of types that can apply to value is restricted to array only.

It also solves __part__ of the type casting below.

Please review this code to make it more "naturally" typescript ready by avoiding using **as** as an "easy way to work around the TS error"

Hugoo (changes requested):

src/lib/decodeData.ts
```diff
@@ -222,6 +223,25 @@ export function decodeKey(schema: ERC725JSONSchema, value) {
...
       }
 
+      if (schema.valueType.includes('[CompactBytesArray]')) {
+        const valueType = schema.valueType.replace('[CompactBytesArray]', '');
+        const valueContent = schema.valueContent.replace(
+          '[CompactBytesArray]',
+          '',
+        );
+
+        if (valueType[0] === '(' && valueType[valueType.length - 1] === ')') {
```
CallumGrindle, line 233: could valueType be something other than an array here?

src/lib/utils.ts
```diff
@@ -282,23 +283,55 @@ export function encodeKey(
...
+          valueType[0] === '(' &&
+          valueType[valueType.length - 1] === ')'
+        ) {
+          const encodedTuples = value.map((element) => {
+            return encodeTupleKeyValue(valueContent, valueType, element);
+          });
+          return encodeValueType('bytes[CompactBytesArray]', encodedTuples);
+        }
+        if (Array.isArray(value) && !Array.isArray(value[0]))
+          return encodeValueType(schema.valueType, value as string[]);
```
CallumGrindle, line 305: could we add some validation somewhere to ensure value is a string array at this point? e.g. here it could be an array of numbers

src/lib/utils.ts
```diff
@@ -282,23 +283,55 @@ export function encodeKey(
...
+            schema.valueType,
+            value as string[] | number[] | JSONURLDataToEncode[],
+          );
       }
 
       return encodeKeyValue(
         schema.valueContent,
         schema.valueType,
-        value,
+        value as
```
CallumGrindle, line 328: can we assert value is not string[][] before this rather than have this big type cast

src/types/ERC725JSONSchema.ts
```diff
@@ -18,20 +18,8 @@ export type ERC725JSONSchemaValueContent =
...
-  | 'bytes4'
-  | 'string[]'
-  | 'address[]'
-  | 'uint256[]'
-  | 'bytes32[]'
-  | 'bytes4[]'
-  | 'bytes[]'
-  | string; // for tuples;
+// https://github.com/lukso-network/LIPs/blob/main/LSPs/LSP-2-ERC725YJSONSchema.md#valueType
+export type ERC725JSONSchemaValueType = string;
```
CallumGrindle, line 22: maybe there is something we could do to separate tuples from other types by having `string: never` or something in ERC725JSONSchemaValueType and a separate type for tuples which is just a string? Then we can still have the security of the type as it was before while allowing tuples to be any string. 

Not too sure how this would look in TS just thinking out loud

src/lib/utils.ts
b00ste: I am not sure how we can do that

src/lib/decodeData.ts
```diff
@@ -111,8 +130,14 @@ export const decodeTupleKeyValue = (
 ) => {
```
samuel-videau, line 130: Can't we have explicit return type?

src/lib/decodeData.ts
```diff
@@ -123,6 +148,7 @@ export const decodeTupleKeyValue = (
     const regexMatch = valueTypePart.match(tupleValueTypesRegex);
 
     if (!regexMatch) {
+      if (valueTypePart === 'address') bytesLengths.push(20);
       return;
```
samuel-videau, line 152: Shouldn't a `else` be added here? I mean the additionnal `if` is useless if after it, it return;

samuel-videau (changes requested):

JeneaVranceanu (changes requested):

src/lib/utils.test.ts
```diff
@@ -132,6 +132,90 @@ describe('utils', () => {
         decodedValue: ['0xcafecafe', '11', '8'],
         encodedValue: '0xcafecafe000000000000000b00000008',
       },
+      {
+        schema: {
+          name: 'TupleMultiType',
+          key: '1e1bc4abe01b7baa7d4a359c0f460e632ef34b3f16f5722bd8892f2dae913022',
+          keyType: 'Singleton',
+          valueType: '(bytes4,address,bytes2)[CompactBytesArray]',
+          valueContent: '(Bytes4,Address,Bytes)',
```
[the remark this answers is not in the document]
CJ42: @JeneaVranceanu `Bytes` is allowed as a `valueContent`. There is not a direct 1 to 1 link between the types in `valueType` and the options provided in `valueContent`.

For instance, the 3rd value provided in the tuple `bytes2` could be interprated as:
- a `Number`
- a `String` (for some text).
- a `Bytes2`
- a `Bytes` (if wanted)

There is no strict rules and direct relations. It is for how it could be interpreted. Interpreting as `Bytes` for any `bytesN` as it is bytes at the end.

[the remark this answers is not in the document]
JeneaVranceanu: Talked about the same with @b00ste just forgot to leave a comment here.
Totally makes sense, it was my mistake. Thanks for the clarifications.

src/lib/utils.ts
```diff
@@ -288,17 +289,52 @@ export function encodeKey(
...
 
+      // This adds an extra check to ensure the casting below is safe
+      // TODO: refactor to fix the TS typing.
+      if (
+        Array.isArray(value) &&
+        Array.isArray(
+          value[0] && !isValidTuple(schema.valueType, schema.valueContent),
+        )
+      ) {
+        throw new Error(`Incorrect value for nested array: not a tuple.`);
```
Hugoo, line 325: pls let's not use backtick when there is no variable in the string:

```suggestion
        throw new Error('Incorrect value for nested array: not a tuple.');
```

src/lib/utils.ts
```diff
@@ -288,17 +289,52 @@ export function encodeKey(
...
         );
       }
 
+      // This adds an extra check to ensure the casting below is safe
+      // TODO: refactor to fix the TS typing.
+      if (
+        Array.isArray(value) &&
+        Array.isArray(
+          value[0] && !isValidTuple(schema.valueType, schema.valueContent),
+        )
```
Hugoo, line 323: why do you put '&& !isValidTuple(schema.valueType, schema.valueContent)' inside the `isArray` ??

Hugoo: I think you want this:

```suggestion
        Array.isArray(value[0]) &&
        !isValidTuple(schema.valueType, schema.valueContent),
```

src/lib/utils.ts
Hugoo: Interesting so, doing it with 2 ifs works:

image

but with one if it doesnt work:

image

I did a few checks tlocally and could manage to make it work.

Let's keep it as it is.

Hugoo (changes requested):

src/lib/decodeData.ts
```diff
@@ -25,11 +25,12 @@ import { isHex } from 'web3-utils';
 import { DecodeDataInput, DecodeDataOutput } from '../types/decodeData';
 import { ERC725JSONSchema } from '../types/ERC725JSONSchema';
 import { isDynamicKeyName } from './encodeKeyName';
-import { valueContentEncodingMap } from './encoder';
+import { valueContentEncodingMap, decodeValueType } from './encoder';
 import { getSchemaElement } from './getSchemaElement';
 import { decodeKeyValue, encodeArrayKey } from './utils';
 
 export const tupleValueTypesRegex = /bytes(\d+)/;
+export const ValueContentsBytesRegex = /Bytes(\d+)/;
```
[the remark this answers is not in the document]
b00ste: The value `tupleValueTypesRegex` was exported before this PR and I think @CJ42 created `ValueContentsBytesRegex` similarly to  `tupleValueTypesRegex`. I will remove the export for both.

src/lib/decodeData.ts
```diff
@@ -222,6 +248,25 @@ export function decodeKey(schema: ERC725JSONSchema, value) {
         );
       }
 
+      if (schema.valueType.includes('[CompactBytesArray]')) {
```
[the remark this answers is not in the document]
b00ste: I created `export const COMPACT_BYTES_ARRAY_STRING = '[CompactBytesArray]';` in the constants file and import it in `decodeData.ts` and `utils.ts`.

Hugoo (approved):

Issue #202, opened 2022-06-28
Add support for `Boolean` valueContent


Hugoo:
https://github.com/lukso-network/LIPs/blob/main/LSPs/LSP-2-ERC725YJSONSchema.md#valueContent

Add the new valueContent in this switch:
https://github.com/ERC725Alliance/erc725.js/blob/develop/src/lib/encoder.ts#L144

Add tests

Hugoo:
@CJ42 can you add more context for this ticket please?

Like what is the input / output for each type - as we can't see it on the specs md

Hugoo:
Edit: `Literal` is already covered ✅

Hugoo:
Zill be closed in. https://github.com/ERC725Alliance/erc725.js/pull/266

CJ42:
Closing as implemented in #266

PR #266, opened 2023-01-31
feat: Add support for `bool` (valueType) and `Boolean` (valueContent)

CJ42:

⭐ Feature

It is not possible to use erc725.js library to:
- encode/decode `bool` or an array of `bool[]` as value types.
- decode data keys that contain `Boolean` in the ERC725Y JSON Schema provided.

- add support for encoding/decoding `bool` and `bool[]` as value type.
- add support for decoding data to be interpreted as `Boolean`

### Other information:

Supersedes + closes #212 
closes issue #202


test/mockSchema.ts
```diff
@@ -789,4 +789,30 @@ export const mockSchema: (ERC725JSONSchema & {
...
+  //     keyType: 'Singleton',
+  //     valueContent: 'Boolean',
+  //     valueType: 'bool',
+  //     // Test data
+  //     returnRawData: abiCoder.encodeParameter('bool', false),
+  //     returnRawDataArray: abiCoder.encodeParameter('bool[]', [false]),
+  //     returnGraphData: false,
+  //     expectedResult: false,
+  //   },
 ];
```
Hugoo, line 818: why this commented code?
These comments are super legacy - i don't see why we would need to add something in this part as a comment?

Hugoo:
@CJ42 the standard mentions `boolean`, not `bool`:
https://github.com/lukso-network/LIPs/blob/main/LSPs/LSP-2-ERC725YJSONSchema.md#valuetype

image

Shall we:

1. Update the standard to **add** `bool` AND add `boolean` support in this PR?
2. Update the standard to **replace** `boolean` to `bool`?
3. Change this PR to support `boolean` instead of `bool` ?

I think 3. is the way to go?

CJ42:
@Hugoo that's a very good point to have spotted! 👌 
I can change it to `boolean` for now.

However, we should either:
- go with option 2) in the future
- allow both `bool` and `boolean`

So in the LSP2 standard, change `boolean` by `bool`. Simply because the `valueType` is for encoding/decoding in the smart contract, according to the smart language (Solidity in our case). 

image

In Solidity, the type is `bool` ✅ (not `boolean` ❌ ).
However☝️ `boolean` is quite common. So maybe we should allow both?

CJ42:
@Hugoo we changed `boolean` by `bool` in LSP2: https://github.com/lukso-network/LIPs/pulls
In the future we should allow both to write either `boolean` or `bool`. `boolean` would fallback to encoding/decoding as `bool`. Opened a feature request: https://github.com/ERC725Alliance/erc725.js/issues/268`

CJ42:
Implemented `boolean` as well. Closes #268

samuel-videau (approved):

src/lib/encoder.ts
```diff
@@ -170,6 +178,10 @@ const valueTypeEncodingMap = {
     encode: (value: string) => abiCoder.encodeParameter('bytes', value),
     decode: (value: string) => abiCoder.decodeParameter('bytes', value),
   },
+  'bool[]': {
+    encode: (value: boolean) => abiCoder.encodeParameter('bool[]', value),
+    decode: (value: string) => abiCoder.decodeParameter('bool[]', value),
+  },
```
Hugoo, line 184: You added `boolean` but I think you would need to also add it here with `boolean[]`

src/types/ERC725JSONSchema.ts
```diff
@@ -31,6 +32,7 @@ export type ERC725JSONSchemaValueType =
   | 'bytes32[]'
   | 'bytes4[]'
   | 'bytes[]'
+  | 'boolean'
```
Hugoo, line 35: so after the discussions, this should be `bool` not `boolean` ?

Hugoo (changes requested):
i think there are still a few missing things

src/types/ERC725JSONSchema.ts
Hugoo: or both as you said you want to support both

src/lib/encoder.ts
CJ42: @Hugoo good catch.
I have added the return type `boolean[]` and some more tests.

src/types/ERC725JSONSchema.ts
CJ42: @Hugoo yes should support both in the library: `bool` and `boolean`. Added the types

Hugoo (approved):

Hugoo (approved):

Issue #280, opened 2023-03-13
Web3 module not found


kalote:
### I'm submitting a...

- [x] bug report
- [ ] feature request
- [ ] question about the decisions made in the repository
- [ ] question about how to use this project

### Summary

When using `"@erc725/erc725.js"` with ethers, it doesn't work as web3 is not present in the project, and it's in the devDependencies of this repository (not included in the final package).

### Other information (e.g. detailed explanation, stack traces, related issues, suggestions how to fix, links for us to have context, eg. StackOverflow, personal fork, etc.)

Stack: NextJS 13, ethers 5.7.2, erc725 0.17.1

Step to reproduce:

- install the package `npm i -S @erc725/erc725.js`
- import and try to use it with ethers (see https://github.com/kalote/manage-up/blob/main/src/context/walletContext.tsx)
- error:

```
./node_modules/@erc725/erc725.js/build/module/src/lib/decodeMappingKey.js:20:0
Module not found: Can't resolve 'web3'
```

Suggestion for fix: Make `web3` module part of the package by moving it in the dependencies (PR incoming)

Hugoo:
Thanks for flagging the issue! Sorry for that!

I sent a patch here: https://github.com/ERC725Alliance/erc725.js/pull/282

PR #281, opened 2023-03-13
fix: adding web3 to dependencies

kalote:

Bug fix

#280 

### Other information:

By including `web3` module in the dependencies, the library will be included in the npm package, preventing the issue #280 .


Hugoo:
Thanks for the PR, to reduce size, let's import from web3-utils rather than adding web3 to the lib, as done here:
https://github.com/ERC725Alliance/erc725.js/pull/282

I will close this PR and merge the one above.

Issue #285, opened 2023-03-16
Add permission to easily check permissions


CallumGrindle:
### I'm submitting a...

- [ ] bug report
- [x] feature request
- [ ] question about the decisions made in the repository
- [ ] question about how to use this project

### Summary
It would be useful to have an easy way to check if a specific subset of permissions are contained in an encoded permission string.

e.g.
`checkPermissions(requiredPermissions: string[] | string, realPermissions: string)`

```js
erc725.checkPermissions(['CHANGEOWNER'], '0x000000000000000000000000000000000000000000000000000000000000ff51')
// true

erc725.checkPermissions(['CHANGEPERMISSIONS', 'CALL'], '0x000000000000000000000000000000000000000000000000000000000000ff51')
// false

erc725.checkPermissions('0x0000000000000000000000000000000000000000000000000000000000000800', '0x000000000000000000000000000000000000000000000000000000000000ff51')
// true
```

PR #286, opened 2023-03-16
feat: Add checkPermissions function

YamenMerhi:

Introduce a new function described in #285 to allow easy detection of which permission is present, supporting plain bytes32 or literal permissions.

### DIFF
--- a/src/lib/detector.ts
+++ b/src/lib/detector.ts
@@ -57,6 +57,11 @@ export const supportsInterface = async (
   }
 };
 
+/**
+ * @notice Check if the given string is a valid 32-byte hex string.
+ * @param str The string to be checked.
+ * @return A boolean value indicating whether the string is a valid 32-byte hex string.
+ */
 function isValid32ByteHexString(str: string): boolean {
   return (
     str.startsWith('0x') &&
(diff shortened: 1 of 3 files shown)


src/lib/detector.ts
```diff
@@ -54,3 +56,32 @@ export const supportsInterface = async (
...
+  grantedPermissions: string,
+): boolean => {
+  // Convert requiredPermissions to an array if it's a single string
+  const requiredPermissionArray: string[] = Array.isArray(requiredPermissions)
+    ? requiredPermissions
+    : [requiredPermissions];
+
+  // Map the literal permissions to their bytes32 representation
+  const mappedPermissionArray: string[] =
+    requiredPermissionArray.map(mapPermission);
```
CallumGrindle, line 75: What if the user passes a string which is not a permission? Will it handle it? Might be useful to add tests for this

YamenMerhi: Added error handling and tests for when it's not a valid bytes32 hex string or known permission.

CallumGrindle (approved):

Hugoo:
Let's not forget to add the documentation for this function + expose it on the main `src/index.ts` file.

PR #287, opened 2023-03-21
refactor!: change `ArrayLength` value from `uint256` to `uint128`

b00ste:
#### ⚠️ BREAKING CHANGES

This PR update how the array length for data of keyType `Array` is encoded.

Since it was changed and reduced from `uint256` to `uint128`.

See --> https://github.com/lukso-network/LIPs/pull/180


b00ste:
This PR is on hold, focusing on other priorities right now.

src/lib/utils.ts
```diff
@@ -259,14 +260,10 @@ export function encodeKey(
         const dataElement = value[index];
         if (index === 0) {
           // This is arrayLength as the first element in the raw array
+          // encoded as a uint128
           results.push({
             key: schema.key,
-            value: encodeKeyValue(
-              'Number',
```
kalote, line 265: So, you're not using `ArrayLength` type here and decided to move to a simpler approach?

src/lib/encoder.ts
```diff
@@ -325,6 +326,10 @@ const valueTypeEncodingMap = {
     decode: (value: string) => abiCoder.decodeParameter('address', value),
   },
   // NOTE: We could add conditional handling of numeric values here...
+  uint128: {
+    encode: (value: string | number) => padLeft(value, 32),
+    decode: (value: string) => toDecimal(value),
+  },
```
Hugoo, line 332: You can't do like this?

```suggestion
  uint128: {
    encode: (value: string | number) => abiCoder.encodeParameter('uint128', value),
    decode: (value: string) => abiCoder.decodeParameter('uint128', value),
  },
```

src/lib/encoder.ts
```diff
@@ -403,10 +408,11 @@ export const valueContentEncodingMap = (valueContent: string) => {
...
     }
     // NOTE: Deprecated. For reference/testing in future
+    // this valueContent does not exist
     case 'ArrayLength': {
       return {
-        type: 'uint256',
-        encode: (value: number | string) => padLeft(numberToHex(value), 64),
+        type: 'uint128',
+        encode: (value: number | string) => padLeft(numberToHex(value), 32),
         decode: (value: string) => hexToNumber(value),
```
Hugoo, line 416: Then let's remove this.

It used to work at a time with uint256.
At no point it time it has or will exist with utin128 as it is already deprecated. I suggest to remove it totally.

src/lib/utils.ts
```diff
@@ -259,14 +260,10 @@ export function encodeKey(
...
-              'uint256',
-              value.length.toString(),
-              schema.name,
-            ) as string,
+            value: leftPad(value.length, 32),
```
Hugoo, line 266: What is the reason for not using the `encodeKeyValue`?

`leftPad(value.length, 32)` is duplicated code with https://github.com/ERC725Alliance/erc725.js/pull/287/files#diff-5927840c3323c66acd20762ba79459d6e5fe967c32d6d2a4be8751fb589c0c93R330

We want only one place in the code that does the encoding, not to sprint it everywhere in the code no?

test/mockSchema.ts
```diff
@@ -226,7 +226,7 @@ export const mockSchema: (ERC725JSONSchema & {
     // testing data
     // the full array of values
     returnRawData: [
-      abiCoder.encodeParameter('bytes', abiCoder.encodeParameter('uint256', 2)), // array length
+      abiCoder.encodeParameter('bytes', abiCoder.encodeParameter('uint128', 2)), // array length
```
Hugoo, line 229: Here you have kept the abiCode.encodeParemeter, nice. Why didn't you update it the same way earlier? cf. the other comment:

https://github.com/ERC725Alliance/erc725.js/pull/287/files#diff-5927840c3323c66acd20762ba79459d6e5fe967c32d6d2a4be8751fb589c0c93R330

Hugoo (commented):
lets clear some stuff before merging - cf comments

src/lib/encoder.ts
CJ42: No this does not work work because  it uses ABI encoding.

ABI encoding will always pas to 32 bytes, even if the type is smaller.

See the Solidity docs:

https://docs.soliditylang.org/en/v0.8.20/abi-spec.html

image

src/lib/utils.ts
CJ42: I think because it was encoding as a valueContent "Number" and the "Number" valueContent in the code uses uint256 as the valueType to encode.

test/mockSchema.ts
CJ42: I am actually not sure about this for the mockSchemas as it is not clear to me.

Here the same reason as before. Because ABI encoding a uint128 still result in a 32 bytes long value.

But we want it to be 16 bytes long.

src/lib/utils.ts
CJ42: No I did not use this because this valueContent is not defined in LSP2.

I guess we might want to deprecate it.

See this discussion comments: https://github.com/ERC725Alliance/erc725.js/pull/287#discussion_r1193203269

src/lib/utils.ts
CJ42: @Hugoo here, after debugging with @kalote we found that `encodeKeyValue` ends up encoding the value as a `Number`, which itself encodes as a `uint256` (and we want to encode Array length as `uint128`.

https://github.com/ERC725Alliance/erc725.js/blob/9e075c4318e0d0055b52b3b70c4ad50ac06c1412/src/lib/encoder.ts#L413-L428

There is a better way, using `encodeValueType` function (already used in this file).

test/mockSchema.ts
CJ42: Ok I was misunderstanding and getting things mixed up. 
I have updated the `mockSchemas` to use `leftPad` the same way as it is done in the library for Array length

src/lib/encoder.ts
Hugoo: Ok I see, nice one, ty for the explanation

kalote (approved):

magalimorin18 (approved):

Hugoo:
@b00ste is it still on hold as described here?
https://github.com/ERC725Alliance/erc725.js/pull/287#issuecomment-1479056978

CJ42:
@Hugoo no it

> @b00ste is it still on hold as described here? [#287 (comment)](https://github.com/ERC725Alliance/erc725.js/pull/287#issuecomment-1479056978)

@Hugoo no it's not on hold anymore. This is ready to be merged once approved.

CallumGrindle (approved):

src/lib/encoder.ts
```diff
@@ -325,6 +326,10 @@ const valueTypeEncodingMap = {
     decode: (value: string) => abiCoder.decodeParameter('address', value),
   },
   // NOTE: We could add conditional handling of numeric values here...
+  uint128: {
+    encode: (value: string | number) => padLeft(value, 32),
```
Hugoo, line 330: what happens if i pass "HelloMyFriend" as an input? It will encode as:

`0x000....000HelloMyFriend` which is not hex and something will fail very hard later in the process no?

What if input is `123456789` ? Will it encode as `0x000...0000123456789` ? Shouldn't it be hex (not decimal) ??

Do we have checks that prevent having corrupted values?

src/lib/utils.test.ts
```diff
@@ -512,7 +510,7 @@ describe('utils', () => {
           '0x3a47ab5bd3a594c3a8995f8fa58d087600000000000000000000000000000000',
         ],
         values: [
-          '0x0000000000000000000000000000000000000000000000000000000000000001',
+          '0x00000000000000000000000000000001',
```
Hugoo, line 513: Can you ad test with values that are wrong? And expect it to throw?

Here we are in a very very happy path.

What if input is "hello world" ? What if input is "11" ?
Do we get `0x00...000011` ? Or should we get `0x00...0000C` ??

Add more test cases please

Hugoo (changes requested):
i have a bad feeling with this encoding. I think it misses a lot of things (if number if > 9, if input is a string?). I feel like we need more validation here. I think the `abiCoder` was doing it for us, but now, you just add 000 on one side :/

src/lib/encoder.ts
```diff
@@ -325,6 +340,10 @@ const valueTypeEncodingMap = {
     decode: (value: string) => abiCoder.decodeParameter('address', value),
   },
   // NOTE: We could add conditional handling of numeric values here...
+  uint128: {
+    encode: (value: string | number) => uint128ToHex(value),
+    decode: (value: string) => hexToUint128(value),
```
Hugoo, line 345: I think you can write the functions here directly instead of making the functions hexToUint128 etc.

src/lib/encoder.ts
```diff
@@ -307,6 +309,19 @@ const decodeStringCompactBytesArray = (compactBytesArray: string): string[] => {
...
 
+const uint128ToHex = (value: string | number) => {
+  const bigNum = BigNumber(value).toString();
+  return '0x' + abiCoder.encodeParameter('uint128', bigNum).substring(34, 66);
+};
+
+const hexToUint128 = (value: string) => {
+  if (!isHex(value)) throw new Error(`${value} is not hex.`);
+  if (value.length > 34)
+    throw new Error(`Too many bytes. ${value.length - 2 / 2} > 16`);
```
Hugoo, line 320:
```suggestion
    throw new Error(`Can't convert hex value ${value} to uint128. Too many bytes. ${value.length - 2 / 2} > 16`);
```

src/lib/encoder.ts
```diff
@@ -307,6 +309,19 @@ const decodeStringCompactBytesArray = (compactBytesArray: string): string[] => {
   return stringValues;
 };
 
+const uint128ToHex = (value: string | number) => {
+  const bigNum = BigNumber(value).toString();
+  return '0x' + abiCoder.encodeParameter('uint128', bigNum).substring(34, 66);
+};
+
+const hexToUint128 = (value: string) => {
+  if (!isHex(value)) throw new Error(`${value} is not hex.`);
```
Hugoo, line 318:
```suggestion
  if (!isHex(value)) throw new Error(`Can't convert ${value} to uint128, value is not hex.`);
```

src/lib/utils.test.ts
```diff
@@ -512,12 +510,67 @@ describe('utils', () => {
...
+          '0xd6c68c2c94af899ce43ff1863693016a711ae7c7',
+          '0x79b698f4bc3051f18b5f94046f09d70823a8fd44',
+          '0x72bebf88546525a5888f188b390701bb0fd9b1a5',
+          '0x882aca051979e32e787e8815d9880759f91e7124',
+          '0x78827c8f8205072858a8cce39b8724d948327ba0',
+          '0xe27cd9c132677cdce2e9efa43b040de35ceff069',
+          '0x072616745957b45c8989e12b9563390fafac4ebe',
+          '0xfd5a7c50c0cf665a772407af3f05522784589c44',
+          '0x13de082cf8a499eee75b0681cfa0141a145f15d9',
+          '0xe3610d0eb167fe7a7b7c25d0aee8874eb8b113ef',
```
Hugoo, line 569: this is copy/pasted from line 525 to 535, put these in a variable and reuse this variable at both places rather than copy/pasting.

Hugoo (approved):

src/lib/encoder.ts
CJ42: I was about to put a review comment for the same thing.

We can probably extract to a function later, we will have to support multiple uintN

src/lib/encoder.ts
```diff
@@ -424,7 +435,7 @@ export const valueContentEncodingMap = (valueContent: string) => {
 
           return padLeft(numberToHex(parsedValue), 64);
         },
-        decode: (value) => '' + hexToNumber(value),
+        decode: (value) => hexToNumber(value).toString(),
```
CJ42, line 438: Why converting it to a string? I think we should decode uintN data as number, not string.

Would suggest to remove this.

@Hugoo @b00ste

b00ste: It was converted to string before, instead of having `'' + hexToNumber(..)` i used the function `toString()`.

CJ42: I removed this behaviour in this PR afterwards.

So I guess we can already replace that here.

https://github.com/ERC725Alliance/erc725.js/pull/289

b00ste: Let's replace it in your next PR, because removing it breaks some tests

src/lib/encoder.ts
```diff
@@ -325,6 +327,22 @@ const valueTypeEncodingMap = {
     decode: (value: string) => abiCoder.decodeParameter('address', value),
   },
   // NOTE: We could add conditional handling of numeric values here...
+  uint128: {
+    encode: (value: string | number) =>
+      '0x' + abiCoder.encodeParameter('uint128', value).substring(34, 66),
+    decode: (value: string) => {
+      if (!isHex(value))
```
CJ42, line 334: Can we put curly brackets for `if` statements, so to keep it consistent with the rest of the code.

Like here:

image

CJ42 (changes requested):

CJ42 (approved):

src/lib/encoder.ts
```diff
@@ -37,8 +37,12 @@ import {
   toChecksumAddress,
   utf8ToHex,
   stripHexPrefix,
+  hexToBytes,
+  bytesToHex,
 } from 'web3-utils';
 
+import BigNumber from 'bignumber.js';
```
Hugoo, line 44: Let's not add this dep: fixed here: https://github.com/ERC725Alliance/erc725.js/pull/304

Issue #277, opened 2023-02-08
Change the way erc725.js encode static `valueType`


CJ42:
### I'm submitting a...

- [x] bug report
- [x] feature request
- [x] question about the decisions made in the repository
- [ ] question about how to use this project

### Summary

Currently, the _erc725.js_ library uses the ABI encoding method from etherjs.
This leads to an encoding format for data that contain padding and offsets, as specified in the [Contract ABI Specs](https://docs.soliditylang.org/en/v0.8.18/abi-spec.html#argument-encoding).

The problem is here is that this encoding format is not valid ❌ 

Because _erc725.js_ is a library used to encode data for storing in the storage of an ERC725Y smart contract, **we should store plain arbitrary bytes, without any formatting or padding** ✅ 

### Detailed informations

We will need to do quite a bit of refactoring in _erc725.js_ for the encoding/decoding of types.
At the moment, for every type, what is being used is the ABI encoding / decoding methods.
This leads to ABI encode, meaning it adds padding + offsets depending on the type (see the screenshot).

Below is a summary of how each `valueType` should be encoded.

| `valueType`  | Encoding  | Example  |
|---|---|---|
| `bool`  | `0x01` (for `true`) or `0x00` (for `false`)  |   |
|  `string` | as utf8 hex bytes <br> **without padding ❌ **  | `"Hello"` --> `0x48656c6c6f`  |
| `address`  | as a 20 bytes long address | `0x388C818CA8B9251b393131C08a736A67ccB19297`  |
| `uin256`  | as a hex value 32 bytes long <br> **left padded** with zeros to fill 32 bytes  |  number `5` --> `0x0000000000000000000000000000000000000000000000000000000000000005`  |
| `uintN` <br> (where N is a multiple of 8 in the range 8 > N > 256) | as a hex value N bytes long <br> **left padded** with zeros to fill N bytes  | number `5` as `uint32` --> `0x00000005`  |
| `bytes32`  | as a hex value 32 bytes long <br> **right padded** to fill 32 bytes  | `0xca5ebeeff00dca11ab1efe6701df563bc1add009076ded6bcf4c6f771f2e3436`  |
| `bytes4`  | as a hex value 4 bytes long <br> **right padded** to fill 4 bytes  | `0xcafecafe`  |
| `bytesN` (from 1 to 32)  | as a hex value N bytes long <br> **right padded** to fill N bytes  |   |
[...]

PR #288, opened 2023-05-12
refactor!: new encoding for static value types (not arrays `[]`)

CJ42:
#### ⚠️ **BREAKING CHANGE**

- **BREAKING CHANGE**: Refactor how _erc725.js_ encode single value types according to the table on the related issue: https://github.com/ERC725Alliance/erc725.js/issues/277

Currently erc725.js add any form of padding and encode smaller types up to 32 bytes words or dynamic types (like `bytes` and `string`) as abi encoded types.

Instead encode as raw bytes with padding as defined by the type (e.g: `uint64` = padded of 8 bytes on the left).

### Other information:

The following changes were also made:

#### 🏗️ Build

Added ethersjs as a dependency in the repo.

#### 🧪 Tests

Re-organise the tests under `encoding.test.ts` to group the value types between simple types and array `[]` types (this is why the diff in tis file is so big). See screenshot below.

image

#### 🐛 Bug Fixes

return number when decoding as plain numbers, not `string`.


package.json
```diff
@@ -79,6 +79,7 @@
   "dependencies": {
     "bignumber.js": "^9.1.1",
     "ethereumjs-util": "^7.1.5",
+    "ethers": "^5.7.2",
```
Hugoo, line 82: Why? Do you understand the impact of this?

With this line, we add the whole ethers lib as a dependency to our lib. This is very likely wayyyy to much things. We want to keep our lib light. At the very least, install "sub libs" - you will notice we didn't install the whole web3js lib but only subparts of it.

I recommend to stick to web3js and not add both web3js and ethers.
The impact is big. Be careful when adding libs to this project. This is not a nodejs backend project where we don't care about the size.

https://bundlephobia.com/package/ethers@6.3.0
https://bundlephobia.com/package/@erc725/erc725.js@0.17.2

test/mockSchema.ts
```diff
@@ -337,7 +337,7 @@ export const mockSchema: (ERC725JSONSchema & {
     ]),
     returnGraphData:
```
Hugoo, line 338: We don't support graphdata anymore, no need to maintain this, you can simply delete.

Hugoo (changes requested):
We can't merge a PR with `wip` commits.
Let's not push a ready to merge PR with wip commits inside.

https://github.com/ERC725Alliance/erc725.js/pull/288/commits/dda2658e9bbe39382a5614598fe05b271048cc8b

image

Hugoo (commented):
this PR is huge and is changing a lot of core tests with a lot of core features. We should be extremely careful.

I would be more confident merging this PR in smaller chunks. Here it really changes the whole core of that lib + tests in one PR. Very dangerous.

Can't approve as it is. Let's discuss the strategy behind this

package.json
CJ42: Ok I understand. I will look at installing only subpackages of ethers related to the encoding only. Not the whole ether.js library.

test/mockSchema.ts
CJ42: @Hugoo 
We need to keep this change otherwise the tests fail. As now we use `hexToNumber` in `encoder.ts` to decode `uintN` numbers.

This function used to return numbers as `string`, but now returns numbers as `number`.
See below when testing on the console of universalprofile.cloud

image

src/lib/encoder.test.ts
```diff
@@ -62,15 +60,12 @@ describe('encoder', () => {
...
-          '0x1337000000000000000000000000000000000000000000000000000000000000',
+        encodedValue: '0x13370000',
       },
       {
         valueType: 'bytes',
-        decodedValue:
-          '0x0000000000000000000000000000000000000000000000000000000000001337',
-        encodedValue:
-          '0x000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000000000000000000000000000000001337',
+        decodedValue: '0x1337',
```
Hugoo, line 67: Maybe we also want to keep that previous test case:

What if the input is: `0x0000000000000000000000000000000000000000000000000000000000001337` ??

Shall we expect this as an output: `0x1337` ?

We are not limited in tests, we can add more and try to go outside the happy path too.

src/lib/encoder.ts
```diff
@@ -309,38 +311,38 @@ const decodeStringCompactBytesArray = (compactBytesArray: string): string[] => {
 
 const valueTypeEncodingMap = {
   bool: {
-    encode: (value: boolean) => abiCoder.encodeParameter('bool', value),
-    decode: (value: string) => abiCoder.decodeParameter('bool', value),
+    encode: (value: boolean) => (value === true ? '0x01' : '0x00'),
```
Hugoo, line 314:
```suggestion
    encode: (value: boolean) => (value ? '0x01' : '0x00'),
```

this should work too

src/lib/encoder.ts
```diff
@@ -309,38 +311,38 @@ const decodeStringCompactBytesArray = (compactBytesArray: string): string[] => {
...
   string: {
-    encode: (value: string) => abiCoder.encodeParameter('string', value),
-    decode: (value: string) => abiCoder.decodeParameter('string', value),
+    encode: (value: string) => utf8ToHex(value),
+    decode: (value: string) => hexToUtf8(value),
   },
   address: {
-    encode: (value: string) => abiCoder.encodeParameter('address', value),
-    decode: (value: string) => abiCoder.decodeParameter('address', value),
+    // TODO: if there is more than 20 bytes, truncate, if there is less than 20 bytes, throw an error
```
Hugoo, line 326: how can an address have more than 20 bytes?

I would not truncate, i would throw an error.

Truncate would make us think yeah it works all good while using another "address string". Let's make it obvious that if length is not 20 then we throw

src/lib/encoder.ts
```diff
@@ -309,38 +311,38 @@ const decodeStringCompactBytesArray = (compactBytesArray: string): string[] => {
...
-      abiCoder.encodeParameter('uint256', value),
-    decode: (value: string) => abiCoder.decodeParameter('uint256', value),
+    encode: (value: string | number) => padLeft(numberToHex(value), 64),
+    decode: (value: string) => hexToNumber(value),
   },
   bytes32: {
-    encode: (value) => abiCoder.encodeParameter('bytes32', value),
-    decode: (value: string) => abiCoder.decodeParameter('bytes32', value),
+    encode: (value) => padRight(value, 64),
+    decode: (value: string) => value.slice(0, 66),
```
Hugoo, line 337: why slice? I tihnk this is confusing.

If the input value here is not bytes32, we throw. We don't "make it work" by slicing it

src/lib/encoder.ts
```diff
@@ -309,38 +311,38 @@ const decodeStringCompactBytesArray = (compactBytesArray: string): string[] => {
...
   },
   bytes4: {
-    encode: (value) => abiCoder.encodeParameter('bytes4', value),
-    decode: (value: string) => abiCoder.decodeParameter('bytes4', value),
+    encode: (value) => padRight(value, 8),
+    decode: (value: string) => value.slice(0, 10),
```
Hugoo, line 341: same here

Also, are we verifying the value here (which we could type too no?) is not too big?

What if input is 0xbeefcafebeef ?

src/lib/encoder.test.ts
Hugoo: Yeah lets add this test case:

```js
{
    valueType: 'bytes4',
    decodedValue: '0x000000000001337',
    encodedValue: '0x1337',
}
```

Might cause issue on decoding test as it is not symetric:

`0x000000000001337` -> Encode -> `0x1337` -> Decode -> `0x1337`

But we can make it work, i think it is nice to have tests like this. (Apply to other types too)

Hugoo: Or... we just throw?

Bc what if:

```
{ valueType: 'bytes4', decodedValue: '0x00000000beef1337', encodedValue: '0x1337', }
```
?

Maybe throwing is safer. As i said in other comment, we don't slice anything. We ask ppl to put the right input

CJ42 (commented):
@Hugoo Thanks for reviewing the PR!

I have replied to some of the comments to discuss the behaviour of the library when data is too small / large compared to its type.

My first feeling would be that:

- when data is smaller than the specified `valueType`, we can pad.

_Example:_

```
data = 0xcafecafe
valueType = bytes10
encoded data = 0xcafecafe000000000000
```

---

Regarding the other way around, when the data is larger than the specified value type, do we:
- throw an error?
- truncate ("cast") the bytes?

```
data = 0xcafecafebeefbeef
valueType = bytes4
encoded data = 0xcafecafe? or throw an error?
```

My option is to throw an error, as there are data lost in the process of the encoding and it is unsafe.

We could provide an `"unsafe"` mode to people as an extra option when encoding, like for example below.

```js
const schemas = [
  {
    name: 'SomeDataKey',
    key: '0xb534ba5bf50f67f445f518a0e2ea5252437868f12b64458a5c624f8ac9d54650',
    keyType: 'Singleton',
    valueType: 'bytes4',
    valueContent: 'Bytes4',
  },
];

myErc725.encodeData(
  [
    {
      keyName: 'SomeDataKey',
      value: '0xcafecafebeefbeef',
    },
  ],
  schemas,
  // optional object of encoding configuration
 { safeEncodingMode: false } // safeEncodingMode is always set to `true` by default in the library when instantiated.
);
/**
{
  keys: ['0xb534ba5bf50f67f445f518a0e2ea5252437868f12b64458a5c624f8ac9d54650'],
  values: ['0xcafecafe]
}
*/
```

> **NB:** the optional flag can be called something else, I just put that as an example. For instance `truncate: true`, or `truncateData: true`.

src/lib/encoder.test.ts
Hugoo: I think if the input is bigger than the type, we should throw. This feels safer imo

src/lib/encoder.ts
Hugoo: @CJ42  I think we should throw. Nothing should be uncertain here.
If you give wrong input, we tell you wrong input. We don't try to make your wrong input work.

src/lib/encoder.ts
Hugoo: I go for throwing an error

Hugoo:
Let's add validation everywhere then, if input data is too big (or not hex or whatever), we throw:

```
Can't encode {inputData} to type {dataType} because {reason}.
```

src/lib/encoder.ts
```diff
@@ -352,21 +366,90 @@ const valueTypeEncodingMap = {
...
+    decode: (value: string) => {
+      if (!isHex(value)) {
+        throw new Error(`Can't convert ${value} to uint256, value is not hex.`);
+      }
+
+      const numberOfBytes = countNumberOfBytes(value);
+
+      if (numberOfBytes > 32) {
+        throw new Error(
+          `Can't convert hex value ${value} to uint256. Too many bytes. ${numberOfBytes} > 32`,
```
magalimorin18, line 381:
```suggestion
          `Can't convert hex value ${value} to uint256. Too many bytes. ${numberOfBytes} is above the maximal number of bytes 32`,
```

src/lib/encoder.ts
```diff
@@ -352,21 +366,90 @@ const valueTypeEncodingMap = {
...
+      } else {
+        valueToEncode = value;
+      }
+
+      const numberOfBytes = countNumberOfBytes(valueToEncode);
+
+      // CHECK that there is not more than 4 bytes
+      if (numberOfBytes > 32) {
+        throw new Error(
+          `Can't convert ${value} to bytes32. Too many bytes, expected at most 32 bytes.`,
```
magalimorin18, line 407:
```suggestion
          `Can't convert ${value} to bytes32. Too many bytes, expected at most 32 bytes, received ${numberOfBytes}.`,
```

src/lib/encoder.ts
```diff
@@ -352,21 +366,90 @@ const valueTypeEncodingMap = {
...
+      } else {
+        valueToEncode = value;
+      }
+
+      const numberOfBytes = countNumberOfBytes(valueToEncode);
+
+      // CHECK that there is not more than 32 bytes
+      if (numberOfBytes > 4) {
+        throw new Error(
+          `Can't convert ${value} to bytes4. Too many bytes, expected at most 4 bytes.`,
```
magalimorin18, line 434:
```suggestion
          `Can't convert ${value} to bytes4. Too many bytes, expected at most 4 bytes, received ${numberOfBytes}.`,
```

magalimorin18 (commented):
Nice to have added all the tests!🪅🎉

src/lib/encoder.test.ts
```diff
@@ -33,329 +33,783 @@ import { JSONURLDataToEncode, URLDataWithHash } from '../types';
...
+          decodedValue: [
+            '0x68114e23B500Cdb63A5B6c9006f3acB0325AD0CC',
+            '0x7466e40FEF4978394A07C9124ad4aD1A374b9465',
+          ],
+          encodedValue:
+            '0x0000000000000000000000000000000000000000000000000000000000000020000000000000000000000000000000000000000000000000000000000000000200000000000000000000000068114e23b500cdb63a5b6c9006f3acb0325ad0cc0000000000000000000000007466e40fef4978394a07c9124ad4ad1a374b9465',
+        },
+        {
+          valueType: 'uint256[]',
+          decodedValue: ['1', '99'], // TODO: return them as an array of `number` type, not an array of `string`
```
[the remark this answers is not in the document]
CJ42: @magalimorin18 this is done in this PR.
https://github.com/ERC725Alliance/erc725.js/pull/289

Basically it means that the decoded values would be returned as:

✅ 
```
[1, 99]
```

not ❌ 

```
['1', '99']
```

src/lib/decodeMappingKey.ts
```diff
@@ -59,7 +55,12 @@ function decodeKeyPart(
         ? 0
         : encodedKeyPart.length - bytesLength;
     decodedKey = encodedKeyPart.slice(sliceFrom);
-  } else decodedKey = decodeValueType(type, make32BytesLong(encodedKeyPart));
+  } else {
+    // this is required if the 2nd word is an address in a MappingWithGrouping
```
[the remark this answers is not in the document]
CJ42: Ok I have fixed that with an additional `if` conditional check --> https://github.com/ERC725Alliance/erc725.js/pull/288/commits/c5d62a25663101bd3b54cd6b7a24a57f68862345

src/lib/encoder.ts
Hugoo: Yes if not address, we throw. We can use `isAddress` from web3: https://web3js.readthedocs.io/en/v1.2.11/web3-utils.html#isaddress

src/lib/encoder.test.ts
```diff
@@ -33,332 +33,783 @@ import { JSONURLDataToEncode, URLDataWithHash } from '../types';
...
-          '0x000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000010000000000000000000000000000000000000000000000000000000000000001',
-      },
-      {
-        valueType: 'boolean[]', // allow to specify "boolean"
-        decodedValue: [false, false],
-        encodedValue:
-          '0x0000000000000000000000000000000000000000000000000000000000000020000000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000',
-      },
-    ];
+    describe('`bool`/`boolean` type', () => {
```
Hugoo, line 36: nice!

src/lib/encoder.test.ts
```diff
@@ -33,329 +33,783 @@ import { JSONURLDataToEncode, URLDataWithHash } from '../types';
...
+          input: '0x000000000001337',
+        },
+        {
+          valueType: 'bytes4',
+          input: '0xcafecafecafecafe',
+        },
+        {
+          valueType: 'bytes4',
+          input: 'hello there', // more than 4 characters
+        },
```
Hugoo, line 117:
```suggestion
        {
          valueType: 'bytes4',
          input: '0x000000000001337', // more than 4 bytes
        },
        {
          valueType: 'bytes4',
          input: '0xcafecafecafecafe', // more than 4 bytes
        },
        {
          valueType: 'bytes4',
          input: 'hello there', // string input that converts to more than 4 bytes in hex
        },
```

src/lib/encoder.ts
```diff
@@ -313,20 +315,32 @@ const decodeStringCompactBytesArray = (compactBytesArray: string): string[] => {
...
+    },
+    decode: (value: string) => hexToUtf8(value),
   },
   address: {
-    encode: (value: string) => abiCoder.encodeParameter('address', value),
-    decode: (value: string) => abiCoder.decodeParameter('address', value),
+    encode: (value: string) => {
+      const abiEncodedValue = abiCoder.encodeParameter('address', value);
+      const bytesArray = hexToBytes(abiEncodedValue);
+      return bytesToHex(bytesArray.slice(12));
```
Hugoo, line 341: why do you need the last 2 lines?

Why do you do hexToBytes then bytesToHex on sliced version?

The abicoder is not enough?

Maybe would be nice to add a comment here to explain why you have to do this. This confuses me.

src/lib/encoder.ts
```diff
@@ -313,20 +315,32 @@ const decodeStringCompactBytesArray = (compactBytesArray: string): string[] => {
...
+    },
+    decode: (value: string) => toChecksumAddress(value),
```
Hugoo, line 343: We had issues with the lib, like if someone writes bad data in the contract, when we decode, it breaks.

Ideally, we want our system to not crash and simply return nothing or skip the value, or maybe flag it as invalid. Without breaking the whole thing.

I think we should make a big check and make sure that errors are properly cought so it doesnt break a "decode all data" function when on of them is corrupted.

src/lib/encoder.ts
```diff
@@ -352,21 +366,90 @@ const valueTypeEncodingMap = {
...
+    decode: (value: string) => {
+      if (!isHex(value)) {
+        throw new Error(`Can't convert ${value} to uint256, value is not hex.`);
+      }
+
+      const numberOfBytes = countNumberOfBytes(value);
+
+      if (numberOfBytes > 32) {
+        throw new Error(
+          `Can't convert hex value ${value} to uint256. Too many bytes. ${numberOfBytes} is above the maximal number of bytes 32`,
```
Hugoo, line 381:
```suggestion
          `Can't convert hex value ${value} to uint256. Too many bytes. ${numberOfBytes} is above the maximal number of bytes 32.`,
```

src/lib/encoder.ts
```diff
@@ -352,21 +366,90 @@ const valueTypeEncodingMap = {
...
+    encode: (value: string | number) => {
+      let valueToEncode: string;
+
+      if (typeof value === 'string' && !isHex(value)) {
+        // if we receive a plain string (e.g: "hey!"), convert it to utf8-hex data
+        valueToEncode = toHex(value);
+      } else if (typeof value === 'number') {
+        // if we receive a number as input, convert it to hex
+        valueToEncode = numberToHex(value);
+      } else {
```
Hugoo, line 398: I think we should add this in the docs. Tell users that if they try to encode a non hex value, we will convert it for them. Maybe you can add an example for each cases:

https://docs.lukso.tech/tools/erc725js/classes/ERC725#encodedata

Here

src/lib/encoder.ts
```diff
@@ -352,21 +366,90 @@ const valueTypeEncodingMap = {
...
+        valueToEncode = value;
+      }
+
+      const numberOfBytes = countNumberOfBytes(valueToEncode);
+
+      // CHECK that there is not more than 4 bytes
+      if (numberOfBytes > 32) {
```
Hugoo, line 405: 4 bytes or 32 bytes?

Hugoo: I think the comment is useless anyway:


this
```
if (numberOfBytes > 32)
```

does not require a comment to tell what it does.

src/lib/encoder.ts
```diff
@@ -352,21 +366,90 @@ const valueTypeEncodingMap = {
...
+      // CHECK that there is not more than 32 bytes
+      if (numberOfBytes > 4) {
+        throw new Error(
+          `Can't convert ${value} to bytes4. Too many bytes, expected at most 4 bytes, received ${numberOfBytes}.`,
+        );
+      }
+
+      const abiEncodedValue = abiCoder.encodeParameter('bytes4', valueToEncode);
+      const bytesArray = hexToBytes(abiEncodedValue);
+      return bytesToHex(bytesArray.slice(0, 4));
```
Hugoo, line 440: This is duplicated code with L390-L411.

I suggest you to make another function with an input param like `encodeTo: 'bytes32' | 'bytes4'` or maybe an int input, and then you reuse this function form both places

Hugoo (commented):
nice tests

src/lib/encoder.ts
Hugoo: DRY

src/lib/encoder.test.ts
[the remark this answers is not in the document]
magalimorin18: ok thanks Jean!

src/lib/encoder.ts
CJ42: Here I did `hexToBytes` and then `bytesToHex` because it is easier to reason in `bytes` in the library.

Otherwise I could do the following screenshot below, but I believe it is much harder to reas and understand because you have to _"calculate in your head"_.

```
0x = 2 characters
12 first bytes = 24 characters

2 + 24 = 26
```

Very hard to understand, I don't like it at all.

src/lib/encoder.ts
CJ42: I will got with the value type input, as number should be only between 1 and 32.

src/lib/encoder.ts
```diff
@@ -83,6 +85,42 @@ const decodeDataSourceWithHash = (value: string): URLDataWithHash => {
   return { hashFunction: hashFunction.name, hash: dataHash, url: dataSource };
 };
 
+const encodeToBytesN = (
```
Hugoo, line 88: nice :)

Hugoo (approved):

PR #289, opened 2023-05-12
fix: decode any `uint256` as `number` ✅ not `string` ❌  + add extra check for literal `valueContent` in tuple decoding

CJ42:
> Awaiting #288 to be merged first

**Bug fix**

There are several TODOs in the TS code, such as:

- currently any `uint256` number being returned (whether standalone or inside an array) is returned as a `string`.
- there is no check for tuples when the `valueContent` of a literal is larger than the `valueType` specified.

This PR fixes these two issues.

### Other information:

N/A


src/lib/decodeData.ts
```diff
@@ -85,12 +86,13 @@ export const isValidTuple = (valueType: string, valueContent: string) => {
       );
     }
 
+    const valueTypeBytesLength = valueTypeParts[i].slice(5);
```
Hugoo, line 89: Where is the 5 coming from?

src/lib/decodeData.ts
```diff
@@ -109,16 +111,25 @@ export const isValidTuple = (valueType: string, valueContent: string) => {
...
 
-    if (
-      valueContentParts[i].slice(0, 2) === '0x' &&
-      !isHex(valueContentParts[i])
-    ) {
-      throw new Error(
-        `Invalid tuple for valueType: ${valueType} / valueContent: ${valueContent}. valueContent of type: ${valueContentParts[i]} is not a valid hex value`,
-      );
-    }
+    const isValueContentHexLiteral = valueContentParts[i].slice(0, 2) === '0x';
```
Hugoo, line 114: Maybe you can use `.startsWith` instead.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/startsWith

src/lib/decodeData.ts
```diff
@@ -109,16 +111,25 @@ export const isValidTuple = (valueType: string, valueContent: string) => {
...
+
+    if (isValueContentHexLiteral) {
+      if (!isHex(valueContentParts[i])) {
```
Hugoo, line 117:
```suggestion
    if (isValueContentHexLiteral && !isHex(valueContentParts[i])) {
```

Hugoo: I think it can be simplified:

```
const isValueContentHexLiteral = valueContentParts[i].slice(0, 2) === '0x';
if (isValueContentHexLiteral) {
      if (!isHex(valueContentParts[i])) {

// ...
```

I think you can do all of this in one line, you can use https://web3js.readthedocs.io/en/v1.2.11/web3-utils.html#ishexstrict -> this will check for 0x AND if it is hexadecimal.

src/lib/encoder.ts
```diff
@@ -518,7 +523,7 @@ export const valueContentEncodingMap = (valueContent: string) => {
 
           return padLeft(numberToHex(parsedValue), 64);
         },
-        decode: (value) => hexToNumber(value).toString(),
+        decode: (value) => hexToNumber(value),
```
Hugoo, line 526: you might need to add `Number` after the PR that was just merged #301

Hugoo (approved):
nice - only this little hex check which looks a bit more complex than it can be imo

src/lib/decodeData.ts
CJ42: The 5th index character in `bytes5` for instance, to extract the `5`.
I have refactored to: 

```js
    const valueTypeBytesLength = valueTypeParts[i].split('bytes')[1];
```

PR #294, opened 2023-07-06
refactor: support `getDataBatch` of 5.0.0 ERC725 release

YamenMerhi:

Support `getDataBatch` of 5.0.0 ERC725 release 
    - Update ERC725Y interfaceId
    - Update function selector for getDataBatch
    - Reference the new version of ERC725Y as `ERC725Y_BATCH` until we remove old version and rename to normal `ERC725Y`

### Other information:

> This is not a cleanup PR, it is intended to make erc725.js able to support the new version of ERC725. A lot of cleanup can happen in later PR, for avoiding duplicated code, changing to the new testsnet, remove support for older version, etc ..

### DIFF
--- a/src/constants/constants.ts
+++ b/src/constants/constants.ts
@@ -34,9 +34,17 @@ export const ERC725Y_INTERFACE_IDS = {
   //     - getData(bytes32[])
   //     - setData(bytes32[],bytes[])
   '3.0': '0x714df77c',
+  // version 5.0.0 removed function overloading
+  // interface functions:
+  //     - getData(bytes32)
+  //     - setData(bytes32,bytes)
+  //     - getDataBatch(bytes32[])
+  //     - setDataBatch(bytes32[],bytes[])
+  '5.0': '0x629aa694',
 };
 
 export enum ERC725_VERSION {
+  ERC725_BATCH = 'ERC725_BATCH', // https://github.com/ERC725Alliance/ERC725/pull/209
   ERC725 = 'ERC725', // https://github.com/ERC725Alliance/ERC725/commit/cca7f98cdf243f1ebf1c0a3ae89b1e46931481b0
   ERC725_LEGACY = 'ERC725_LEGACY',
   NOT_ERC725 = 'NOT_ERC725',
@@ -52,8 +60,16 @@ export const METHODS: Record<Method, MethodData> = {
     returnEncoding: Encoding.BYTES,
   },
   [Method.GET_DATA]: {
-    // https://github.com/ERC725Alliance/erc725/blob/main/docs/ERC-725.md#erc725y
+    // https://github.com/ERC725Alliance/ERC725/blob/v4.0.0/docs/ERC-725.md#erc725y
     sig: '0x4e3e6e9c',
+    gas: numberToHex(2000000),
+    gasPrice: numberToHex(100000000),
+    value: numberToHex(0),
+    returnEncoding: Encoding.BYTES_ARRAY,
+  },
+  [Method.GET_DATA_BATCH]: {
+    // https://github.com/ERC725Alliance/ERC725/blob/v5.1.0/docs/ERC-725.md#erc725y
+    sig: '0xdedff9c6',
     gas: numberToHex(2000000),
     gasPrice: numberToHex(100000000),
     value: numberToHex(0),
--- a/src/constants/constants.ts
+++ b/src/constants/constants.ts
@@ -34,6 +34,7 @@ export const ERC725Y_INTERFACE_IDS = {
   //     - getData(bytes32[])
   //     - setData(bytes32[],bytes[])
   '3.0': '0x714df77c',
+  // InterfaceId of version 3 == interfaceId of version 4
   // version 5.0.0 removed function overloading
   // interface functions:
   //     - getData(bytes32)
--- a/src/constants/constants.ts
+++ b/src/constants/constants.ts
@@ -49,7 +49,7 @@ export enum ERC725_VERSION {
   // The ERC725Y_LEGACY version uses getData(bytes32) function
   ERC725_LEGACY = 'ERC725_LEGACY',
   // The ERC725_v2 version uses getData(bytes32[]) function, as well as v3 and v4
-  ERC725_v2 = 'ERC725', // https://github.com/ERC725Alliance/ERC725/releases/tag/v2.2.0
+  ERC725_v2 = 'ERC725_v2', // https://github.com/ERC725Alliance/ERC725/releases/tag/v2.2.0
   // The ERC725_v5 version uses getDataBatch(bytes32[]) function
   ERC725_v5 = 'ERC725_v5', // https://github.com/ERC725Alliance/ERC725/releases/tag/v5.0.0
 }
--- a/src/constants/constants.ts
+++ b/src/constants/constants.ts
@@ -45,10 +45,13 @@ export const ERC725Y_INTERFACE_IDS = {
 };
 
 export enum ERC725_VERSION {
-  ERC725_BATCH = 'ERC725_BATCH', // https://github.com/ERC725Alliance/ERC725/pull/209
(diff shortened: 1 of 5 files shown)


magalimorin18 (approved):

magalimorin18:
lgtm

src/constants/constants.ts
```diff
@@ -34,9 +34,17 @@ export const ERC725Y_INTERFACE_IDS = {
   //     - getData(bytes32[])
   //     - setData(bytes32[],bytes[])
   '3.0': '0x714df77c',
+  // version 5.0.0 removed function overloading
```
CJ42, line 37: @Hugoo should we also add version 4?
From what I remember, we mention that we don't want to support old versioning. So it might be worth thinking if we should just delete the old versions mentioned above.

CJ42: Also can't we import the @erc725/smart-contracts package and import the ERC725Y interface ID of v5 from the package?

Can we install @erc725/smart-contracts as a dev dependency to prevent having the package size to be too big, or we can't have it as dev-dependency?

src/index.test.ts
```diff
@@ -304,6 +304,55 @@ describe('Running @erc725/erc725.js tests...', () => {
...
+          url: 'ipfs://QmecrGejUQVXpW4zS948pNvcnQrJ1KiAoM6bdfrVcWZsn5',
+        },
+      },
+      {
+        name: 'LSP1UniversalReceiverDelegate',
+        key: '0x0cfc51aec37c55a4d0b1a65c6255c4bf2fbdf6277f3cc0730c45b828b6db8b47',
+        value: '0x36e4Eb6Ee168EF54B1E8e850ACBE51045214B313',
+      },
+    ];
+    it('with web3.currentProvider [ERC725Y_BATCH]', async () => {
```
CJ42, line 345: Let's add a space before the it("") test

CJ42 (approved):
LGTM. Added few questions in comment that can be discussed after

src/constants/constants.ts
YamenMerhi: @CJ42 In terms of `interfaceId`, version 3 is the same as version 4. That's why it's not added.
For deleting the old versions, at some point, they should be deleted, but I guess this should not be done in this PR nor in the near future unless we reach a stable version that goes to production.

src/index.test.ts
```diff
@@ -304,6 +304,56 @@ describe('Running @erc725/erc725.js tests...', () => {
...
+        },
+      },
+      {
+        name: 'LSP1UniversalReceiverDelegate',
+        key: '0x0cfc51aec37c55a4d0b1a65c6255c4bf2fbdf6277f3cc0730c45b828b6db8b47',
+        value: '0x36e4Eb6Ee168EF54B1E8e850ACBE51045214B313',
+      },
+    ];
+
+    it('with web3.currentProvider [ERC725Y_BATCH]', async () => {
```
CallumGrindle, line 346: are there other providers we can add testcases for?

CallumGrindle (approved):

src/index.test.ts
Hugoo: i tihnk no, we used to have apollo / graphql but not anymore.

src/constants/constants.ts
YamenMerhi: @Hugoo Done 🙏

src/constants/constants.ts
```diff
@@ -52,13 +60,21 @@ export const METHODS: Record<Method, MethodData> = {
...
-    // https://github.com/ERC725Alliance/erc725/blob/main/docs/ERC-725.md#erc725y
+    // https://github.com/ERC725Alliance/ERC725/blob/v4.0.0/docs/ERC-725.md#erc725y
     sig: '0x4e3e6e9c',
     gas: numberToHex(2000000),
     gasPrice: numberToHex(100000000),
     value: numberToHex(0),
     returnEncoding: Encoding.BYTES_ARRAY,
   },
+  [Method.GET_DATA_BATCH]: {
+    // https://github.com/ERC725Alliance/ERC725/blob/v5.1.0/docs/ERC-725.md#erc725y
```
Hugoo, line 72: nice

src/constants/constants.ts
```diff
@@ -52,13 +60,21 @@ export const METHODS: Record<Method, MethodData> = {
...
+    sig: '0xdedff9c6',
+    gas: numberToHex(2000000),
+    gasPrice: numberToHex(100000000),
```
Hugoo, line 75: I think we will need to make it dynamic at some point no? cc @CallumGrindle @magalimorin18

src/index.test.ts
```diff
@@ -304,6 +304,56 @@ describe('Running @erc725/erc725.js tests...', () => {
     });
   });
 
+  describe('Getting data (using new getDataBatch) in schema by provider [e2e] - luksoTestnet', () => {
+    const web3 = new Web3('https://rpc.testnet.lukso.network');
+    const ERC725_BATCH_CONTRACT_ADDRESS =
+      '0x4b30900F119E11D2A8CAe18176c4f9840E586Cc4';
+
+    const e2eSchema: any = [
```
Hugoo, line 312: Why do you use `any` as a type? 

Ideally, we should never use any. It should be like very rare case and exceptions. Let's not take the habit of writing any too much.

src/provider/providerWrapper.ts
```diff
@@ -61,6 +61,15 @@ export class ProviderWrapper {
...
 
   async getErc725YVersion(address: string): Promise<ERC725_VERSION> {
+    const isErc725YBatch = await this.supportsInterface(
+      address,
+      ERC725Y_INTERFACE_IDS['5.0'],
+    );
+
+    if (isErc725YBatch) {
+      return ERC725_VERSION.ERC725_BATCH;
+    }
```
Hugoo, line 71: similar to my comment above, i would stick to version number instead of features.

src/index.test.ts
```diff
@@ -304,6 +304,56 @@ describe('Running @erc725/erc725.js tests...', () => {
     });
   });
 
+  describe('Getting data (using new getDataBatch) in schema by provider [e2e] - luksoTestnet', () => {
```
Hugoo, line 307: is there a non e2e test? A test with mock provider?

src/provider/providerWrapper.ts
```diff
@@ -206,6 +217,47 @@ export class ProviderWrapper {
...
+      Method.GET_DATA_BATCH,
+      results[0].result,
+    );
+
+    return keyHashes.map<GetDataReturn>((key, index) => ({
+      key,
+      value: decodedValues ? decodedValues[index] : decodedValues,
+    }));
+  }
+
```
Hugoo, line 260: I would reco to not create this funciton which is 95% the same as the function below.

I bet there were a few copy pastes here.

image

More code more problems, and please DRY ;)
It makes it hard to see what is the diff between batch and non batch, i used the diff checker to see that the only difference is that you use `Method.GET_DATA_BATCH` instead of `Method.GET_DATA`.

So, you add 40 lines of code to just replace 3 variables.

I suggest:

```
getData(address: string, keyHashes: string, useBatchMethod: boolean) {
 // ...
 const decodedValues = decodeResult(
      useBatchMethod ? Method.GET_DATA_BATCH : Method.GET_DATA,
      results[0].result,
    );
// ...
}
```

And you can also add comment here to see that getData was replaced by getDataBatch from version 5

Hugoo (changes requested):

src/constants/constants.ts
CallumGrindle: whats this being used for? The lib doesn't send any transactions does it?

src/provider/providerWrapper.ts
```diff
@@ -197,29 +207,36 @@ export class ProviderWrapper {
...
       default:
         return [];
     }
   }
 
-  private async _getAllData(
+  private async _getAllDataGeneric(
     address: string,
     keyHashes: string[],
+    method: Method,
```
Hugoo, line 228: Very cool! You could even be more precise here, as we really need only 2 methods.

Hugoo (approved):

PR #300, opened 2023-07-31
docs: use toggles to display better examples of encoding

CJ42:
In the docs, all the examples are one after the other, making the code snippets very long and big.

- Sort them between toggles that can be opened, so that it is easier to find the relevant example for the developer.
- Improve subheadings so that examples can be more easily found from the sidebar

Screenshot 2023-07-31 at 16 35 51

image


magalimorin18:
super cool !

Hugoo (approved):

PR #301, opened 2023-08-01
chore: update web3 to v1.10.0 and prettier to v3

Hugoo:


package.json
```diff
@@ -64,23 +64,23 @@
...
+    "web3": "^1.10.0"
   },
   "dependencies": {
     "bignumber.js": "^9.1.1",
     "ethereumjs-util": "^7.1.5",
-    "web3-eth-abi": "^1.8.2",
-    "web3-providers-http": "^1.8.0",
-    "web3-utils": "^1.8.2"
+    "web3-eth-abi": "^1.10.0",
+    "web3-providers-http": "^1.10.0",
```
CJ42, line 83: Nice! There were some improvements in this release

https://github.com/web3/web3.js/releases/tag/v1.10.0


image

src/index.ts
```diff
@@ -430,7 +430,7 @@ export class ERC725 {
   static encodePermissions(permissions: Permissions): string {
     const result = Object.keys(permissions).reduce((previous, key) => {
       return permissions[key]
-        ? previous + hexToNumber(LSP6_DEFAULT_PERMISSIONS[key])
+        ? previous + Number(hexToNumber(LSP6_DEFAULT_PERMISSIONS[key]))
```
CJ42, line 433: Why is this casting necessary? Doesn't the function `hexToNumber` convert to a `Number` already?

CJ42 (approved):

CallumGrindle (approved):

src/index.ts
Hugoo: The type has changed in v1.9.0 -> https://github.com/web3/web3.js/releases/tag/v1.9.0

> utils.toNumber and utils.hexToNumber can now return the large unsafe numbers as BigInt, if true was passed to a new optional parameter called bigIntOnOverflow (https://github.com/web3/web3.js/pull/5845)

We dont call it with true so it should not output a BigInt, but still the TS type thinks it is Number or string. This is to limit to Number only (which should be the truth)

PR #302, opened 2023-08-02
fix: dependencies & example

kalote:

- readme update with testnet
- `example/` folder update

### Other information:

the `example/` folder now contains my UP address as example. I'm fine with it, but we might want to setup a 'showcase' UP address to put instead. UP to you (🤣)!

### DIFF
--- a/examples/package.json
+++ b/examples/package.json
@@ -9,8 +9,8 @@
     "decodeData": "node src/decodeData.js"
   },
   "dependencies": {
-    "@erc725/erc725.js": "^0.14.0-beta.0",
-    "web3": "4.0.2"
+    "@erc725/erc725.js": "^0.17.2",
+    "web3": "^1.10.0"
   },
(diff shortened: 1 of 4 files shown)


Hugoo (approved):

examples/package.json
```diff
@@ -9,8 +9,8 @@
     "decodeData": "node src/decodeData.js"
   },
   "dependencies": {
-    "@erc725/erc725.js": "^0.14.0-beta.0",
-    "web3": "4.0.2"
+    "@erc725/erc725.js": "^0.17.2",
+    "web3": "^1.10.0"
```
Hugoo, line 13: Why downgrade web3 ? I think ultimately we will want to have v4 ? (with ts support etc.)

kalote: I wanted to keep the same version between `example` folder and the root package

PR #309, opened 2023-09-14
fix: Remove hardcoded require statement inside of module (esm) code.

richtera:

There was a

```js
/* eslint-disable-next-line */
const HttpProvider = require("web3-providers-http");
```

in the code which is illegal in a esm module.

When using a true esm compiler like esbuild or bun the require function will no longer be defined
in the esm (module) context.

It will work in cjs and esm

### Other information:

https://app.clickup.com/t/2645698/DEV-8268

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -21,6 +21,7 @@
  */
 
 import { hexToNumber, isAddress, leftPad, toHex } from 'web3-utils';
+import HttpProvider from 'web3-providers-http';
 
 import { ProviderWrapper } from './provider/providerWrapper';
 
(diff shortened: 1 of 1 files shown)


JeneaVranceanu (approved):

Hugoo (commented):
wow - interesting - i guess there was a reason ?

weird

Hugoo (approved):

Hugoo:
Waiting for @skimaharvey answer in our chat before merging

skimaharvey (approved):

PR #316, opened 2023-10-17
feat: add permission `EXECUTE_RELAY_CALL` + use Bitwise OR `|` to toggle ON permissions

CJ42:


magalimorin18 (approved):

CallumGrindle (commented):
what about the permission value for the encode/decode permissions function?

CallumGrindle (approved):

PR #317, opened 2023-10-17
refactor!: deprecate static / non-static `supportsInterface` methods

CJ42:

**Deprecation**

The _erc725.js_ library contains a list of interface IDs related to the `@lukso/lsp-smart-contracts`. These interface IDs are out of date and never maintain, and **should not be in this package**, but instead imported as:

```
import { INTERFACE_IDS } from "@lukso/lsp-smart-contracts";
```

And simply checked using a ERC165 ABI

```js
  try {
    const eip165Contract = contract(eip165ABI as any, address)
    return await eip165Contract.methods.supportsInterface(interfaceId).call()
  } catch (error: unknown) {
    console.error(error)
    return false
  }
```

Deprecate the function `supportsInterface` from the _erc725.js_ library, to avoid future errors and maintenance for a function that is not used and should not be used.


CallumGrindle (approved):

frozeman (changes requested):
Lets please discuss, as this is valid feature but should not work like solidly but rather:
```
erc725js.supporstInterface('ERC725X')
erc725js.supporstInterface('LSP0')
...
```

dapp developers need to be able to check with contract they are dealing with

Hugoo:
It currently works with both - interfaceId (hex) and name.

image


I agree this might belong to another package as this is not really erc725 related.

We have no rush to remove that and I think we use this feature in other apps. Maybe let's wait until it is avail. in an other lib - such as lsp-utils before removing it from here?

Hugoo:
I close because it is a duplicate of: https://github.com/ERC725Alliance/erc725.js/pull/352

PR #325, opened 2023-10-24
feat!: add `encode/decodeValueType` as public callable methods

CJ42:

Feature

The methods `encodeValueType` and `decodeValueType` are currently not available as public method part of the ERC725 class.

These methods can be useful to encode certain LSP2 value type raw. For example, a `uin128` (For LSP2 Array length) to obtain an exactly 16 bytes long value.

Make this method part of the class, so that they can be called and used.


src/index.ts
```diff
@@ -168,6 +169,21 @@ export class ERC725 {
 
     throw new Error(`Incorrect or unsupported provider ${providerOrRpcUrl}`);
   }
+
+  private getAddressAndProvider() {
```
CJ42, line 173: @Hugoo I just moved this function at the top, to group the `private` methods together.

frozeman:
we need to update the docs as well, when we do this change

src/index.ts
Hugoo: ok nice ty

Hugoo:
Ready to merge - just waiting for conflict fix and docs update: https://github.com/ERC725Alliance/erc725.js/blob/develop/docs/classes/ERC725.md

docs/classes/ERC725.md
```diff
@@ -766,86 +975,6 @@ myErc725.decodeMappingKey(
 
 ---
 
-## encodePermissions
```
CJ42, line 769: @Hugoo this has not been deleted, just move at the top to sort function names correctly by alphabetical order.

PR #326, opened 2023-10-25
feat: allow to encode LSP2 Array length only

CJ42:

feature (wip)

Currently it is only possible to encode a full array of valueType when the keyType is `Array`. 

Allow to only encode the Array length only of `uint128` that is 16 bytes long, not a full array.


src/lib/utils.test.ts
```diff
@@ -405,6 +407,21 @@ describe('utils', () => {
...
+        keyType: 'Array',
+        valueContent: 'Address',
+        valueType: 'address',
+      };
+
+      const decodedValue = 3;
+      const encodedValue = '0x00000000000000000000000000000003';
+
+      assert.equal(encodeKey(schema, decodedValue), encodedValue);
+    });
```
Hugoo, line 424: what is the reason for putting this `it` under the `describe('encodeArrayKey',` describe?

The function being tested is `encodeKey` so I suggest to put it under the corresponding "describe" (prob `describe('encodeKey/decodeKey', () => {`)

Hugoo (approved):

PR #332, opened 2023-10-31
Feat/gas param

### LINKED ISSUE #328: Weird forced gas limit on getData
https://github.com/ERC725Alliance/erc725.js/blob/9f1b5fdd0cb4eceab4396fb2d4b7fd02a8a17d2a/src/constants/constants.ts#L69

This hard codes the gas limit to 2M for retrieving a data item (or various other flavors of the ERC725Y methods). This seems arbitrary and unreasonable, especially as some contracts might want to dynamically generate data items. Unless there's some very good reason, would it be possible to remove that limitation?

kalote:
New `gas` configuration in the optional `params` setting:
```
const config = {
  ipfsGateway: 'https://YOUR-IPFS-GATEWAY/ipfs/',
  gas: 20_000_000, // optional, default is 2_000_000
};

const erc725 = new ERC725(schemas, address, RPC_URL, config);
```

I also removed the `gasPrice` from the constants.ts file as it was unused.

Fix #328

### DIFF
--- a/docs/classes/ERC725.md
+++ b/docs/classes/ERC725.md
@@ -1446,7 +1446,7 @@ On non instantiated class, you should provide an `options` object.
 | :-------- | :----- | :------------------------------------------------------------------- |
 | `address` | string | Address of the smart contract to check against a certain interface.  |
 | `rpcUrl`  | string | RPC URL to connect to the network the smart contract is deployed to. |
-| `gas`     | number | Optional: gas parameter to use. Default: 2_000_000.                  |
+| `gas`     | number | Optional: gas parameter to use. Default: 1_000_000.                  |
 
 #### Returns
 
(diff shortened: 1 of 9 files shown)


karalabe:
I don't know anything about ts, so I've no idea what and how it works :D... but I would also recommend supporting setting the gas to 0 as a wildcard (or null or something) that would remove the call limit completely and let the node run with whatever cap it has itself.

richtera (approved):

skimaharvey:
> ### What kind of change does this PR introduce (bug fix, feature, docs update, ...)?
> New `gas` configuration in the optional `params` setting:
> 
> ```
> const config = {
>   ipfsGateway: 'https://YOUR-IPFS-GATEWAY/ipfs/',
>   gas: 20_000_000, // optional, default is 2_000_000
> };
> 
> const erc725 = new ERC725(schemas, address, RPC_URL, config);
> ```
> 
> I also removed the `gasPrice` from the constants.ts file as it was unused.
> 
> ### What is the current behaviour (you can also link to an open issue here)?
> Fix #328
> 
> ### What is the new behaviour (if this is a feature change)?
> ### Other information:

Also agree. `2_000_000` is so arbitrary. May be would exceed the gas block limit for some chains or would be irrelevant for others.

kalote:
So shall we set the default to 0? @CallumGrindle / @skimaharvey / @Hugoo ?

kalote:
Default is 2_000_000 but can be change to 0 if needed via the param. I guess it can be added to the next release @CallumGrindle

kalote:
Default gas value set to 1_000_000 (1Gwei).

PR #343, opened 2023-11-06
fix: Change to verification object

richtera:

Change from verificationData/Function or hash/hashFunction to verification object.
Fix chai assert and debug jests.
Repair getVerificationMethod (getVerificationFunction) that said it would match signature or name
but only matched on signature.

The current values were recently changed from hash/hashFunction to verificationData/verificationFunction
but we also want a source property. Therefore this changes it to an object containing
{ method, source?, data }?

All jests pass so it should not introduce functionality changes other than the property values.

### Other information:

https://github.com/lukso-network/LIPs/blob/main/LSPs/LSP-3-Profile-Metadata.md


examples/src/instantiation.js
```diff
@@ -61,67 +61,75 @@ export const profileJson = {
       {
         width: 1800,
         height: 1712,
-        verificationFunction: 'keccak256(bytes)',
-        verificationData:
-          '0xfbcfcbbc86d886e862419361c48251d778884a90429d36c8d002559cbcb52972',
+        verification: {
+          method: 'keccak256(bytes)',
```
CJ42, line 65: Why here do we use `keccak256(bytes)`?

While in the previous file `example/src/getData.js` we use `keccak256(utf8)`?

image

Not sure as both are related to a LSP3Profile. I guess one of the example might be wrong.

package.json
```diff
@@ -77,7 +77,9 @@
     "web3": "^1.10.0"
   },
   "dependencies": {
+    "@types/jest": "^29.5.7",
```
CJ42, line 80: Shouldn't this be in the dev dependencies? This might increase the size of the package and not necessary to ship with the package.
@Hugoo ?

src/constants/constants.ts
```diff
@@ -92,22 +92,22 @@ export const METHODS: Record<Method, MethodData> = {
...
   KECCAK256_BYTES = 'keccak256(bytes)',
 }
 
-export enum SUPPORTED_VERIFICATION_FUNCTION_HASHES {
+export enum SUPPORTED_VERIFICATION_METHOD_HASHES {
   HASH_KECCAK256_UTF8 = '0x6f357c6a',
   HASH_KECCAK256_BYTES = '0x8019f9b1',
 }
 
 export type SUPPORTED_VERIFICATION_FUNCTIONS =
```
CJ42, line 105: Shouldn't this be changed to below for consistency?

```suggestion
export type SUPPORTED_VERIFICATION_METHODS =
```

src/index.test.ts
```diff
@@ -54,26 +53,19 @@ const address = '0x0c03fba782b07bcf810deb3b7f0595024a444f4e';
...
-        new ERC725(mockSchema, address, { test: false });
-      },
-      (error: any) =>
-        error.message.indexOf('Incorrect or unsupported provider') >= -1,
-    );
+    assert.throws(() => {
+      // @ts-ignore
+      // eslint-disable-next-line no-new
+      new ERC725(mockSchema, address, { test: false });
+    }, /Incorrect or unsupported provider/);
```
CJ42, line 68: NIce it's more readable.
Can't we put the error as a string tho?

```suggestion
    }, "Incorrect or unsupported provider");
```

src/index.test.ts
```diff
@@ -680,8 +674,10 @@ describe('Running @erc725/erc725.js tests...', () => {
 
                 // Encoded value of:
                 // {
-                //   verificationFunction: 'keccak256(bytes)', // 0x8019f9b1
-                //   verificationData: '0xc41589e7559804ea4a2080dad19d876a024ccb05117835447d72ce08c1d020ec',
+                //   verification: {
+                //     method: 'keccak256(bytes)', // 0x8019f9b1
```
CJ42, line 678: Shouldn't it be 

```suggestion
                //     method: 'keccak256(utf8)', // 0x8019f9b1
```

src/lib/encoder.ts
```diff
@@ -44,14 +44,11 @@ import {
   toBN,
 } from 'web3-utils';
 
-import { JSONURLDataToEncode, URLDataWithHash } from '../types';
+import { JSONURLDataToEncode, URLDataWithHash, Verification } from '../types';
```
CJ42, line 47: What is this type? Not clear when reading it

src/lib/encoder.ts
```diff
@@ -60,30 +57,33 @@ const bytesNRegex = /Bytes(\d+)/;
...
-
+  if (!verification || !verification.method || !verification.data) {
+    throw new TypeError('No verification function or data provided.');
+  }
+  const verificationMethod = getVerificationMethod(verification.method);
   return (
-    keccak256(verificationFunction.name).slice(0, 10) +
-    dataHash.slice(2) +
+    keccak256(verificationMethod.name).slice(0, 10) +
+    verification.data.slice(2) +
```
CJ42, line 69: Suggest using method from web3.js for readability

```suggestion
    stripHexPrefix(verification.data) +
```

src/types/encodeData/JSONURL.ts
```diff
@@ -9,15 +9,19 @@ interface URLData {
   url: string;
 }
 
+export interface Verification {
```
CJ42, line 12: Shouldn't we rename this to something a bit more specific? Like

```suggestion
export interface VerificationOptions {
```

or

```suggestion
export interface VerificationInput {
```

examples/src/instantiation.js
richtera: I think it's because for LSP3Profile it refers to the utf8 JSON data.

package.json
richtera: Moving it, makes sense.

richtera: FYI: jest was also missing, causing the repo to assume jest from global. The current version of jest is 29 which is most of the time not compatible with 27.

src/constants/constants.ts
richtera: Changing.

examples/src/instantiation.js
richtera: The keccak256(bytes) is used for images for example.

src/lib/encoder.ts
richtera: ok

src/lib/encoder.ts
richtera: It's the JSON version of the JSONURL, Hmmm, this is not new though.

src/types/encodeData/JSONURL.ts
richtera: Well it's the verification structure, so I think it's clearest if it doesn't have an appendage Options or Info or anything, but that's completely subjective.

package.json
richtera: Moved.

src/index.test.ts
richtera: I suppose we could. I just converted the original code to the new one leaving the same expression.

src/index.test.ts
richtera: I think it's supposed to be the keccak of the byte stream, but you have a good point. The IPFS cid is already a hash, but I guess not for other URLs types. So keeping a hash of just the URL is probably not good verification.

src/lib/encoder.ts
```diff
@@ -60,30 +57,33 @@ const bytesNRegex = /Bytes(\d+)/;
...
 const encodeDataSourceWithHash = (
-  hashType: SUPPORTED_VERIFICATION_FUNCTIONS,
-  dataHash: string,
+  verification: undefined | Verification,
   dataSource: string,
 ): string => {
-  const verificationFunction = getVerificationFunction(hashType);
-
+  if (!verification || !verification.method || !verification.data) {
+    throw new TypeError('No verification function or data provided.');
```
CallumGrindle, line 64: do we definitely want to throw here?

src/types/decodeData.ts
```diff
@@ -15,13 +15,14 @@ export interface DecodeDataInput extends DataInput {
...
 export interface DecodeDataOutput {
-  value: string | string[] | URLDataWithHash;
+  value: string | string[] | URLDataWithHash | null;
   name: string;
   key: string;
 }
 
 export interface FetchDataOutput {
   value:
+    | null
```
CallumGrindle, line 25: just to clarify why do we return null here now?

CallumGrindle (commented):
thank you for these changes @richtera 🙏

examples/src/instantiation.js
CJ42: Double checked this and that's correct 👍 

image

image

src/index.test.ts
CJ42: Interesting

src/types/encodeData/JSONURL.ts
CJ42: Ok let's keep it like this then

CJ42 (approved):

src/lib/encoder.ts
richtera: Changed it to support a 0000 or unknown verification method. This will always verify as good.

src/types/decodeData.ts
richtera: We need the null for the jests. We were not typechecking fully previously. We use null in places although it should really be `value?` but I didn't think that makes sense to mix into this PR.

CallumGrindle: nice thanks for clarifying

CallumGrindle (approved):

PR #353, opened 2023-11-09
refactor: improve code for tuple checks

CJ42:

**Refactor**

Current code base work fine, but not easy to understand the code behind the tuple logic.

Improve the code for tuple checking by create internal methods to valid and extract tuple content.

### DIFF
--- a/src/lib/decodeData.ts
+++ b/src/lib/decodeData.ts
@@ -19,7 +19,6 @@
  * @author Callum Grindle <@CallumGrindle>
  * @date 2023
  */
-
 import { isHexStrict } from 'web3-utils';
 import { COMPACT_BYTES_ARRAY_STRING } from '../constants/constants';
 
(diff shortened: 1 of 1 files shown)


src/lib/decodeData.ts
```diff
@@ -139,27 +141,19 @@ export const decodeTupleKeyValue = (
...
   const bytesLengths: number[] = [];
+
   valueTypeParts.forEach((valueTypePart) => {
     const regexMatch = valueTypePart.match(tupleValueTypesRegex);
 
     // if we are dealing with `bytesN`
-    if (regexMatch) {
-      bytesLengths.push(parseInt(regexMatch[1], 10));
-    }
+    if (regexMatch) bytesLengths.push(parseInt(regexMatch[1], 10));
```
Hugoo, line 156: personal preference - but i prefer with brackets, i think it is easier to read and see it is an if

src/lib/decodeData.ts
```diff
@@ -33,36 +32,39 @@ import { decodeKeyValue, encodeArrayKey } from './utils';
...
     return false;
   }
 
-  // At this stage, we can assume the user is trying to use a tuple, let's throw errors instead of returning
-  // false
+  return true;
+};
 
-  let valueTypeToDecode = valueType;
+const extractTupleElements = (tupleContent: string): string[] =>
```
Hugoo, line 49: nice

Hugoo (approved):
nice

PR #355, opened 2023-11-20
feat: add support to encode / decode any `uint8` to `uint256` types

CJ42:

**New feature**

Currently, erc725.js allows to encode only numbers `valueType` of `uint128` or `uint256`.
Any other type available in the 8 bit range from Solidity are not available.

Refactor the value type map to filter by regex and allow to encode number from `uint8` to `uint256`.

### DIFF
--- a/src/lib/encoder.ts
+++ b/src/lib/encoder.ts
@@ -51,12 +51,17 @@ import {
   SUPPORTED_VERIFICATION_METHOD_STRINGS,
   UNKNOWN_VERIFICATION_METHOD,
 } from '../constants/constants';
-import { getVerificationMethod, hashData, countNumberOfBytes } from './utils';
+import {
+  getVerificationMethod,
+  hashData,
+  countNumberOfBytes,
+  isValidUintSize,
+} from './utils';
 import { ERC725JSONSchemaValueType } from '../types/ERC725JSONSchema';
 
 const abiCoder = AbiCoder;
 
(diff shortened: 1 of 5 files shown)


src/lib/encoder.ts
```diff
@@ -356,156 +359,354 @@ const decodeStringCompactBytesArray = (compactBytesArray: string): string[] => {
...
+//       // if we receive a number as input,
+//       // convert each letter to its utf8 hex representation
+//       if (typeof value === 'number') {
+//         return utf8ToHex(`${value}`);
+//       }
+
+//       return utf8ToHex(value);
+//     },
+//     decode: (value: string) => hexToUtf8(value),
+//   },
```
Hugoo, line 574: what is this big commented block?

src/lib/encoder.ts
```diff
@@ -52,10 +52,13 @@ import {
   UNKNOWN_VERIFICATION_METHOD,
 } from '../constants/constants';
 import { getVerificationMethod, hashData, countNumberOfBytes } from './utils';
+import { ERC725JSONSchemaValueType } from '../types/ERC725JSONSchema';
 
 const abiCoder = AbiCoder;
 
-const bytesNRegex = /Bytes(\d+)/;
+const uintNValueTypeRegex = /^uint(?:8|[1-9]\d|1[0-9]{2}|2[0-4]\d|25[0-6])$/;
```
Hugoo, line 59: I have doubts over this one

- it is hard to read
- it catches values which should not be ok (uint23...)

I suggest to use just a catch to get N and then test N with proper maths

Hugoo (approved):

PR #357, opened 2023-11-23
feat: add support for multi dynamic key name

### LINKED ISSUE #356: [Bug]: `encodeKeyName` does not work for LSP8 keys that contain multiple types in dynamic key parts
### I'm submitting a...

- [x] bug report
- [ ] feature request
- [ ] question about the decisions made in the repository
- [ ] question about how to use this project

### Summary

The following code fails for the data keys `LSP8MetadataTokenURI`.

image

The following error get returned.

image

Hugoo:
DEV-9056

https://app.clickup.com/t/2645698/DEV-9056

--

Closes #356


dzbo (approved):

CallumGrindle (approved):
lgtm, though Im not sure I fully understand how the `duplicateMultiTypeERC725SchemaEntry` works. 
We create schemas for each possible typing, and progress as if each one was passed individually. How are we sure that the correct type is used for encoding/decoding? I may lack context

Hugoo:
> lgtm, though Im not sure I fully understand how the `duplicateMultiTypeERC725SchemaEntry` works. We create schemas for each possible typing, and progress as if each one was passed individually. How are we sure that the correct type is used for encoding/decoding? I may lack context

The correct type is given by the user. The user will use the lib like this:

image

And because of this "duplicate step" the lib will be able to encode/decode properly.

PR #367, opened 2023-12-12
refactor!: update new LSP7/8 interface IDs

CJ42:


magalimorin18 (approved):

Hugoo:
pplease link to source ref so we can easily check

src/constants/interfaces.ts
```diff
@@ -21,8 +21,8 @@ export const INTERFACE_IDS_0_12_0 = {
   LSP1UniversalReceiver: '0x6bb56a14',
   LSP1UniversalReceiverDelegate: '0xa245bbda',
   LSP6KeyManager: '0x23f34c62',
-  LSP7DigitalAsset: '0xdaa746b7',
-  LSP8IdentifiableDigitalAsset: '0x30dc5278',
+  LSP7DigitalAsset: '0xb3c4928f',
+  LSP8IdentifiableDigitalAsset: '0xecad9f75',
```
Hugoo, line 25: Value doesn't match with:

https://docs.lukso.tech/contracts/interface-ids

I guess the docs are wrong?

cc @YamenMerhi

Hugoo: I read `0x3a271706` in

https://github.com/lukso-network/lsp-smart-contracts/blob/develop/constants.ts#L33

Hugoo: Changed last week in:

https://github.com/lukso-network/lsp-smart-contracts/commit/2bf84f638926a9969385ce12f98f389b26ca0173

YamenMerhi: The correct value is `0x3a271706`, https://github.com/lukso-network/lsp-smart-contracts/blob/lsp-smart-contracts-v0.14.0/constants.ts#L33

PR #373, opened 2024-01-05
fix: Add workaround to read (bytes4,URI), repair unknown signature to pass.

richtera:

Add workaround to read old (bytes4,URI) as a VerifiableURI without verification.
Allow 0x00000000 verification everywhere and don't convert it to unknown
Pass unknown verification signatures back.

Both 0x00000000 and unknown turn into "unknown" making it impossible to create third party verification methods while using erc725.js
We were unable to read the old (bytes4,URI) for LSP8TokenMetadataBaseURI

LSP8TokenMetadataBaseURI is converted and returns as VerifiableURI information.
Saving and retrieving an unknown verification methods is now possible with getData
The 0x00000000 method is now defined as a constant as well.

### DIFF
--- a/src/lib/encoder.ts
+++ b/src/lib/encoder.ts
@@ -136,7 +136,7 @@ const decodeDataSourceWithHash = (value: string): URLDataWithHash => {
 
   try {
     const dataSource = hexToUtf8('0x' + encodedData); // Get as URI
-    if (encodedData.length < 64 || /^[a-z]{2,}:/.test(dataSource)) {
+    if (encodedData.length < 64 || /^[a-z]{2,}:[/\S]/.test(dataSource)) {
       return {
         verification: {
           method: NONE_VERIFICATION_METHOD,
--- a/src/lib/encoder.ts
+++ b/src/lib/encoder.ts
@@ -136,8 +136,7 @@ const decodeDataSourceWithHash = (value: string): URLDataWithHash => {
 
   try {
     const dataSource = hexToUtf8('0x' + encodedData); // Get as URI
-    if (encodedData.length < 64 || /^(https?|ipfs):\/\//.test(dataSource)) {
-      console.log(dataSource);
+    if (encodedData.length < 64 || /^[a-z]{2,}:/.test(dataSource)) {
       return {
         verification: {
           method: NONE_VERIFICATION_METHOD,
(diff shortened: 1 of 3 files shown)


src/lib/encoder.ts
```diff
@@ -129,12 +133,33 @@ const decodeDataSourceWithHash = (value: string): URLDataWithHash => {
   const verificationMethodSignature = value.slice(0, 10);
   const verificationMethod = getVerificationMethod(verificationMethodSignature);
   const encodedData = value.slice(10); // Rest of data string after function hash
+
+  try {
+    const dataSource = hexToUtf8('0x' + encodedData); // Get as URI
+    if (encodedData.length < 64 || /^(https?|ipfs):\/\//.test(dataSource)) {
+      console.log(dataSource);
```
CallumGrindle, line 140: can remove this log

src/lib/encoder.ts
```diff
@@ -129,12 +133,33 @@ const decodeDataSourceWithHash = (value: string): URLDataWithHash => {
...
+
   const dataHash = '0x' + encodedData.slice(0, 64); // Get jsonHash 32 bytes
   const dataSource = hexToUtf8('0x' + encodedData.slice(64)); // Get remainder as URI
 
   return {
     verification: {
-      method: verificationMethod?.name || UNKNOWN_VERIFICATION_METHOD,
+      method:
+        verificationMethod?.name ||
+        (verificationMethodSignature === '0x00000000'
```
CallumGrindle, line 160:
```suggestion
        (verificationMethodSignature === NONE_VERIFICATION_METHOD
```

CallumGrindle (approved):
we should also add some comments in the code explaining why its doing this. I think it looks okay, question is once its in will we remove it or just allow this for the future?

src/lib/encoder.ts
Hugoo: veirifacrionMethodSignature === "0x0000..." will return a bool right? Do we want a book here in method?

Hugoo (commented):
I agree with Callum - not a fan of introducing quick hacks. Maybe worth to have comments everywhere to explain this has been done like that.

src/lib/encoder.ts
richtera: It is now, you're right. Initially the NONE_VERIFICATION_METHOD was an empty string, but it turns out that caused problems elsewhere. Fixing.

richtera: Simplified to just return the verificationMethodSignature.
The "unknown" thing was what was causing problems because as a user of erc725 it was not possible to determine what the signature was for a custom verification method and 0x0000000 would also be marked as unknown.

PR #374, opened 2024-01-10
feat: add permission related to 4337

CJ42:


skimaharvey (approved):

YamenMerhi (approved):

docs/classes/ERC725.md
```diff
@@ -754,7 +754,8 @@ ERC725.encodePermissions({
   ENCRYPT: false,
   DECRYPT: false,
   SIGN: false,
-  EXECUTE_RELAY_CALL: false
+  EXECUTE_RELAY_CALL: false,
+  "4337_PERMISSION": false
```
richtera, line 758: It's not good practice to have property names start with numeric. ERC4337_PERMISSION would be a lot better to deal with in javascript because it's a standard identifier.

richtera (commented):
Probably use ERC4337_PERMISSION and not just 4337_PERMISSION because without the ERC prefix it's not a legitimate identifier.

richtera (approved):

PR #375, opened 2024-01-15
fix: Make data retrieval more resilient. Do not throw on ipfs data problems.

richtera:

- fix: Repair problems with IPFS, fetch and VerifiableURI
- fix: Repair to not throw errors when data is not authentica or not accessible withing getDataFromExternalSources.

The current behavior is that when data is invalid or unreachable in ipfs. getDataFromExternalSource will throw an exception.

These are now converted to a console.error, allowing the data to return as "null". This allows getData to be called with multiple keys
including one that maybe invalid and not just throw out of the section of code.

### Other information:

I also made a similar kind of fix when the array length data (i.e. LSP12IssuedAssets[]) contains a uint256 instead of a uint128.

### DIFF
--- a/src/index.test.ts
+++ b/src/index.test.ts
@@ -569,7 +569,6 @@ describe('Running @erc725/erc725.js tests...', () => {
           ),
         );
         const result = await erc725.fetchData('TestJSONURL');
-        console.log(fetchStub.getCalls(), result);
         fetchStub.restore();
 
         assert.deepStrictEqual(result, {
--- a/src/index.test.ts
+++ b/src/index.test.ts
@@ -561,8 +561,15 @@ describe('Running @erc725/erc725.js tests...', () => {
         const jsonString = `{"LSP3Profile":{"profileImage":"ipfs://QmYo8yg4zzmdu26NSvtsoKeU5oVR6h2ohmoa2Cx5i91mPf","backgroundImage":"ipfs://QmZF5pxDJcB8eVvCd74rsXBFXhWL3S1XR5tty2cy1a58Ew","description":"Beautiful clothing that doesn't cost the Earth. A sustainable designer based in London Patrick works with brand partners to refocus on systemic change centred around creative education. "}}`;
 
         const fetchStub = sinon.stub(global, 'fetch');
-        fetchStub.onCall(0).returns(Promise.resolve(new Response(jsonString)));
+        fetchStub.onCall(0).returns(
+          Promise.resolve(
+            new Response(jsonString, {
+              headers: { 'content-type': 'application/json' },
+            }),
+          ),
+        );
         const result = await erc725.fetchData('TestJSONURL');
+        console.log(fetchStub.getCalls(), result);
         fetchStub.restore();
 
         assert.deepStrictEqual(result, {
@@ -584,7 +591,7 @@ describe('Running @erc725/erc725.js tests...', () => {
           [contractVersion.interface],
         );
 
-        const ipfsGateway = 'https://2eff.lukso.dev';
+        const ipfsGateway = 'https://api.universalprofile.cloud';
 
         const erc725 = new ERC725([testJSONURLSchema], address, provider, {
           ipfsGateway,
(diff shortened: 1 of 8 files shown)


src/index.test.ts
```diff
@@ -561,8 +561,15 @@ describe('Running @erc725/erc725.js tests...', () => {
...
-        fetchStub.onCall(0).returns(Promise.resolve(new Response(jsonString)));
+        fetchStub.onCall(0).returns(
+          Promise.resolve(
+            new Response(jsonString, {
+              headers: { 'content-type': 'application/json' },
+            }),
+          ),
+        );
         const result = await erc725.fetchData('TestJSONURL');
+        console.log(fetchStub.getCalls(), result);
```
CallumGrindle, line 572: think we can remove this log :)

src/lib/encoder.test.ts
```diff
@@ -666,13 +667,15 @@ describe('encoder', () => {
...
-      it('throws when trying to decode a bytes17 as `uint128`', () => {
-        expect(() =>
-          decodeValueType('uint128', '0x000000000000000000000000000000ffff'),
-        ).to.throw(
-          "Can't convert hex value 0x000000000000000000000000000000ffff to uint128. Too many bytes. 17 > 16",
-        );
-      });
+      // We do not want this during decode. During encode we enforce the correct sizes.
+      //
+      // it('throws when trying to decode a bytes17 as `uint128`', () => {
```
CallumGrindle, line 672: maybe we can adjust this test so that it no longer throws

src/lib/encoder.ts
```diff
@@ -72,7 +72,7 @@ const encodeDataSourceWithHash = (
   dataSource: string,
 ): string => {
   const verificationMethod = getVerificationMethod(
-    verification?.method || UNKNOWN_VERIFICATION_METHOD,
+    verification?.method || '0x00000000',
```
CallumGrindle, line 75: we can use the NONE_VERIFICATION_METHOD

src/index.test.ts
richtera: Oh I missed that one. Yea some of the test scripts needed debugging.

src/lib/encoder.test.ts
richtera: Ah good call.

CallumGrindle (approved):
left some comments. Looks okay to me

richtera:
Rebased.

src/lib/encoder.ts
```diff
@@ -529,11 +532,12 @@ const valueTypeEncodingMap = (
...
           }
 
+          // Although this is "wrong" making this a real Exception prevents people from using the library
+          // rather than encouraging them to fix the problem. Making this a console error.
           const numberOfBytes = countNumberOfBytes(value);
-
           if (numberOfBytes > (uintLength as number) / 8) {
-            throw new Error(
-              `Can't convert hex value ${value} to ${type}. Too many bytes. ${numberOfBytes} > 16`,
+            console.error(
```
CJ42, line 539: I think this is incorrect and this part might need to be refactored. If there is more bytes than `number of bits / 8`, it should be fine as long as the left most bytes are `00`.

If for instance for a `uint8`, there is more than 1 byte, that's ok as long as the left most byte is `00`. 

For instance, to decoding 0x00fe as uint8 would be ok, it will decode as 254 ✅ 
But if you try to decode 0x01fe as uint8, it should not allow you ❌

src/lib/encoder.ts
```diff
@@ -782,7 +786,7 @@ export const valueContentEncodingMap = (
         },
         decode: (value: string) => {
           if (typeof value !== 'string' || !isHex(value)) {
-            console.log(`Value: ${value} is not hex.`);
+            console.error(`Value: ${value} is not hex.`);
```
CJ42, line 789: Nice idea to make it error in the console!

CJ42 (commented):
Added some review comments

Issue #372, opened 2024-01-04
[Bug] Not possible to fetch dynamic key parts individually


fhildeb:
### Issue

Dynamic key parts of ERC725Y storage can only be fetched when I provide more than keys:

```js
let data = await erc725js.fetchData([
  'LSP4TokenName',
  {
    keyName: 'LSP4CreatorsMap:<address>',
    dynamicKeyParts: '0x9139def55c73c12bcda9c44f12326686e3948634',
  },
]);
```

However, fetching **only one key** does not work for any dynamic objects it seems.

a) It neither works with a single key

```js
let data = await erc725js.fetchData({
  keyName: 'LSP4CreatorsMap:<address>',
  dynamicKeyParts: '0x9139def55c73c12bcda9c44f12326686e3948634',
});
```

b) Nor an array with a single key:

```js
let data = await erc725js.fetchData([{
  keyName: 'LSP4CreatorsMap:<address>',
  dynamicKeyParts: '0x9139def55c73c12bcda9c44f12326686e3948634',
}]);
```

### Why is this a problem

- No unified `fetchData` or `getData` calls, causing problems in automation
- It is not possible to retrieve data if there is **only one key**

### What I would like to know

Could someone explain 

- why this decision was made like that?
- if there is a workaround?

### Related Links 

- [Throwing in `src/lib/decodeData.ts`]( https://github.com/ERC725Alliance/erc725.js/blob/330c604f55601b634f0eea381bc542e72021d19c/src/lib/decodeData.ts#L161)
- [Related Documentation](https://github.com/ERC725Alliance/erc725.js/blob/develop/docs/classes/ERC725.md#single-key-example-1)

richtera:
One thing I did notice is that for a single item (not array) the return is also not an array, but that should not affect a single key array input when looking at getData.
For a single key it will return the value also without array notation.
```js
  const data = await getDataMultiple(erc725Options, [keyOrKeys]);

  return data[0];
```
So I believe since fetchData uses getData instead of getDataMultiple it's not getting an array in the single item case. But it doesn't explain the single array input case.

CJ42:
Tested with an other data key like `AddressPermissions:Permissions:<address>` works.

```js
let lsp6Result = await erc725js.getData([{ keyName: 'AddressPermissions:Permissions:<address>', dynamicKeyParts: '0x9139def55c73c12bcda9c44f12326686e3948634' }]);
```

[...]

PR #378, opened 2024-01-16
fix: Add bit count level verification.

richtera:

Add bit count level verification and errors inside of encode/decode for u?int[0-9]\*

Currently it lets some things pass by to be flexible and doesn't detect actual overflow situations.

This will error if the number of bits don't fit and warn if the number of bytes are larger than expected.


src/lib/utils.ts
```diff
@@ -588,6 +589,16 @@ export function countNumberOfBytes(data: string) {
   return stripHexPrefix(data).length / 2;
 }
 
+export function countSignificantBits(data = '0x0') {
+  const bytes = hexToBytes(data);
+  for (let i = 0; i < bytes.length; i++) {
+    if (bytes[i] !== 0) {
+      return (bytes.length - i - 1) * 8 + 32 - Math.clz32(bytes[i]);
```
CJ42, line 596: What does this Math function do? It seems quite hard to understand this line of code.

src/lib/utils.ts
```diff
@@ -588,6 +589,16 @@ export function countNumberOfBytes(data: string) {
   return stripHexPrefix(data).length / 2;
 }
 
+export function countSignificantBits(data = '0x0') {
```
CJ42, line 592: Why giving a default value here? Also isn't should have at least two nibbles to be a valid byte?

```suggestion
export function countSignificantBits(data = '0x00') {
```

src/lib/utils.ts
CJ42: I am not sure to understand this line of code.

src/lib/decodeData.ts
```diff
@@ -312,37 +312,33 @@ export function decodeData(
...
-      schemaElement = getSchemaElement(schema, keyName, dynamicKeyParts);
-
-      // NOTE: it might be confusing to use as the output will contain other keys as the ones used
-      // for the input
-      return {
-        key: schemaElement.key,
-        name: schemaElement.name,
-        value: decodeKey(schemaElement, value),
-      };
+    const schemaElement: ERC725JSONSchema = isDynamic
```
CJ42, line 321: Why this line was changed? Is it related to the bit count?
Feels like it might related to issue #372 or am I completely wrong?

src/lib/utils.ts
richtera: It calculates the number of zero bits if the byte value had 32 bits. So the value for 0 is 32 and the value for 1 is 31 and for 2 is 30, 4 it's 29 and so on. The whole expression then returns the number of non zero bits.

src/lib/utils.ts
richtera: The reason was that otherwise the code couldn't be tested using javascript in node. I'll change it back.

richtera: Removed.

src/lib/decodeData.ts
richtera: It just converts the expression using a tenary early so that it was easier to create a common exception handler.

src/lib/getDataFromExternalSources.ts
```diff
@@ -51,53 +54,92 @@ export const getDataFromExternalSources = (
...
-    let receivedData;
-    try {
-      const { url } = patchIPFSUrlsIfApplicable(urlDataWithHash, ipfsGateway);
+      const urlDataWithHash: URLDataWithHash =
+        dataEntry.value as URLDataWithHash; // Type URLDataWithHash
 
-      receivedData = await fetch(url).then(async (response) => {
-        if (!response.ok) {
-          return undefined;
+      const { url } = patchIPFSUrlsIfApplicable(
```
CallumGrindle, line 78: could this destructuring fail?

CallumGrindle (approved):
lgtm

src/lib/getDataFromExternalSources.ts
richtera: Checking, but I don't think so.

richtera: No unless the data is already null ```js
export function patchIPFSUrlsIfApplicable(receivedData, ipfsGateway) {
    if (receivedData &&
        receivedData.url &&
        receivedData.url.indexOf('ipfs://') !== -1) {
        return Object.assign(Object.assign({}, receivedData), { url: receivedData.url.replace('ipfs://', ipfsGateway) });
    }
    return receivedData;
}
```. Not this is kind of problematic until the the data-providers url resolver is implemented to support arweave and other URLs. Having an ipfs:// constant inside of the code will be difficult to support going forward.

PR #379, opened 2024-01-17
docs: Add TypeScript to getting started page

fhildeb:
### What kind of change does this PR introduce?

Update the docs

### What is the current behavior?

Currently throwing errors for TS users that are unaware of built-in types.

### What is the new behavior?

They instantiate their schemas with the `ERC725JSONSchema[]` type.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,44 +41,85 @@ If you install it on the backend side, you may need to also install [`isomorphic
 
 ## Instantiation
 
-You need to initialise it with a [schema](https://docs.lukso.tech/tools/erc725js/schemas), a contract address and an RPC URL.
+You need to initialise the ERC725 object with a [schema](https://docs.lukso.tech/tools/erc725js/schemas), a contract address, and an RPC URL.
 
 ```js
-import { ERC725 } from '@erc725/erc725.js';
+const address = '0x0Dc07C77985fE31996Ed612F568eb441afe5768D';
+const RPC_URL = 'https://rpc.testnet.lukso.network';
+const config = {
+  ipfsGateway: 'https://YOUR-IPFS-GATEWAY/ipfs/',
+  gas: 20_000_000, // optional, default is 1_000_000
+};
+```
 
-// Part of LSP3-Profile Schema
-// https://github.com/lukso-network/LIPs/blob/master/LSPs/LSP-3-Profile-Metadata.md
-const schema = [
+### TypeScript
+
+> If you are using ES6 `import` statements in Node.js, make sure your file has a `.mjs` extension, or that your project is set up to support ES6 modules.
+
+```ts
+import { ERC725, ERC725JSONSchema } from '@erc725/erc725.js';
+
+// Part of LSP3-UniversalProfile Schema
+// https://github.com/lukso-network/LIPs/blob/master/LSPs/LSP-3-UniversalProfile.md
+const schemas: ERC725JSONSchema[] = [
   {
     name: 'SupportedStandards:LSP3Profile',
     key: '0xeafec4d89fa9619884b600005ef83ad9559033e6e941db7d7c495acdce616347',
     keyType: 'Mapping',
-    valueContent: '0x5ef83ad9',
     valueType: 'bytes',
+    valueContent: '0x5ef83ad9',
   },
   {
     name: 'LSP3Profile',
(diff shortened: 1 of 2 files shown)


docs/getting-started.md
```diff
@@ -27,7 +27,7 @@ If you install it on the backend side, you may need to also install [`isomorphic
 
 ## Instantiation
 
-```js
+```ts
```
richtera, line 30: Might be cool to have two sections one ts and one js. People usually look for one if they don't know typescript they won't know to remove the while ` as ERC725JSONSchema[]`. Or add a commented out line "use this for plain javascript". Also NOTE: import is only available inside of a "mjs" file when running it locally in node. So the "plain version" should probably also use "require" or specify that it has to be ".mjs".

richtera (commented):
Probably not a big deal, but taking editing typecript for someone looking for plain javascript might not be awesome.

docs/getting-started.md
```diff
@@ -63,9 +63,11 @@ const config = {
   gas: 20_000_000, // optional, default is 1_000_000
 };
 
-const erc725 = new ERC725(schemas, address, RPC_URL, config);
+const erc725 = new ERC725(schemas as ERC725JSONSchema[], address, RPC_URL, config);
 ```
 
+> If you are using plain JavaScript, you can remove the `ERC725JSONSchema[]` type.
```
CJ42, line 69: Suggest using tabs instead of this note, as suggested by @richtera

CJ42 (changes requested):

docs/getting-started.md
CJ42: Or then no tabs but two code snippets examples

CJ42:
I would suggest actually to not go with this change in the docs and instead fix the typing at the schema levels.

Ideally, the schemas should be exported from an `index.ts` file under the `schema/` folder, and have the right typing assigned to them, as:

```ts
export const LSP6Schema: ERC725JSONSchema[] = [
    {
        // schema here
    },
    // more objects
]
```

richtera:
But aren't the schemas usually in json files? Making them ts would be better for sure

docs/getting-started.md
```diff
@@ -62,6 +60,22 @@ const config = {
...
+```
+
+### TypeScript
+
+> If you are using ES6 `import` statements in Node.js, make sure your file has a `.mjs` extension, or that your project is set up to support ES6 modules.
+
+```ts
+import { ERC725 } from '@erc725/erc725.js';
+
+const erc725 = new ERC725(schemas as ERC725JSONSchema[], address, RPC_URL, config);
```
Hugoo, line 72:
```suggestion
import { ERC725, ERC725JSONSchema } from '@erc725/erc725.js';

const erc725 = new ERC725(schemas as ERC725JSONSchema[], address, RPC_URL, config);
```

Hugoo (approved):

docs/getting-started.md
```diff
@@ -28,11 +28,26 @@ If you install it on the backend side, you may need to also install [`isomorphic
...
-import { ERC725 } from '@erc725/erc725.js';
+const address = '0x0Dc07C77985fE31996Ed612F568eb441afe5768D';
+const RPC_URL = 'https://rpc.testnet.lukso.network';
+const config = {
+  ipfsGateway: 'https://YOUR-IPFS-GATEWAY/ipfs/',
+  gas: 20_000_000, // optional, default is 1_000_000
+};
+```
+
+> All used schemas should be exported from an `index.ts` file under the `schema/` folder.
```
CJ42: Please delete this line

CJ42 (approved):

README.md
```diff
@@ -41,44 +41,85 @@ If you install it on the backend side, you may need to also install [`isomorphic
...
-const config = {
-  ipfsGateway: 'https://2eff.lukso.dev/ipfs/',
-};
+const erc725 = new ERC725(schemas, address, RPC_URL, config);
+```
+
+### JavaScript
+
+```js
+import { ERC725 } require('@erc725/erc725.js');
```
Hugoo, line 94: this is weird

Hugoo (commented):
I don't understand why we should duplicate the whole code snippet
just to explain that ppl can use `: ERC725JSONSchema[]`

I would reco to keep only one code snippet and put a comment just where needed

// Typescript type: : ERC725JSONSchema[]

or smtgh

Just to let ppl super easily spot the difference and to have less stuff to maintain

PR #380, opened 2024-01-22
fix: decoding tuple when no value is fetched

### LINKED ISSUE #372: [Bug] Not possible to fetch dynamic key parts individually

CJ42:

Bug fix

See issue #372 

When using `getData` or `fetchData` with a key type that contains a tuple, if no value is fetched, return an array with a number of `undefined` elements (as many as defined by the number of elements in the tuple).

### Other information:

Fix #372

### DIFF
--- a/src/index.test.ts
+++ b/src/index.test.ts
@@ -720,6 +720,56 @@ describe('Running @erc725/erc725.js tests...', () => {
           );
         });
       }
+
+      it('fetchData and return `[undefined...]` number of elements based on number of tuple elements', async () => {
+        const provider = new HttpProvider(
+          {
+            returnData: allRawData.filter(
+              (rawData) =>
+                rawData.key ===
+                '0xd154e1e44d32870ff5ade9e8726fd06d0ed6c996f5946dabfdfd46aa6dd2ea99',
+            ),
+          },
+          [contractVersion.interface],
+        );
+
+        const erc725js = new ERC725(
+          [
+            {
+              name: 'LSP4CreatorsMap:<address>',
+              key: '0x6de85eaf5d982b4e5da00000<address>',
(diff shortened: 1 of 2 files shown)


CJ42:
Closing as this was already resolved in #378

PR #384, opened 2024-02-22
feat: add ALL_PERMISSIONS flag to encoding

skimaharvey:
# What does this PR introduce? 

- [x] Introduce the `ALL_PERMISSIONS` flag to encoding

Let's you set all permissions on one flag. Also added the logic where you can do combinations where:

 - `ALL_PERMISSIONS` with `CHANGE_OWNER` gives you the right hex. 
 - `ALL_PERMISSIONS` with permissions that are already part of the `ALL_PERMISSIONS` value does not affect the hex etc


src/constants/constants.ts
```diff
@@ -179,7 +179,8 @@ export const LSP6_DEFAULT_PERMISSIONS = {
   DECRYPT                          : "0x0000000000000000000000000000000000000000000000000000000000100000",
   SIGN                             : "0x0000000000000000000000000000000000000000000000000000000000200000",
   EXECUTE_RELAY_CALL               : "0x0000000000000000000000000000000000000000000000000000000000400000",
-  ERC4337_PERMISSION               : "0x0000000000000000000000000000000000000000000000000000000000800000"
+  ERC4337_PERMISSION               : "0x0000000000000000000000000000000000000000000000000000000000800000",
```
CJ42, line 182: Suggest adding a comment that this is not a standard permission defined in LSP6, but used for integration of ERC4337 standard.

Otherwise, people might look into the LSP6 specs to understand what this is used for, but will not find it.

Do we have a link in our docs or a guide that mentions this permission? Maybe worth adding it here.

@Hugoo @skimaharvey wdyt?

src/index.test.ts
```diff
@@ -1337,6 +1343,7 @@ describe('Running @erc725/erc725.js tests...', () => {
           SIGN: false,
           EXECUTE_RELAY_CALL: false,
           ERC4337_PERMISSION: false,
+          ALL_PERMISSIONS: false,
```
CJ42, line 1346: It might be worth adding some extra tests to tests that when all permission is set and others are set to false, it removes them, etc... More cases

CJ42 (commented):
Added 1st round of review

src/index.test.ts
skimaharvey: yes ideally we can randomize them like i did for some tests i added. I find it not efficient to do one test at a time. Will think about it

src/constants/constants.ts
skimaharvey: i agree that this is confusing. I am not sure if the confusion starts at the lib level but imo more at the LIPs level and probably we should fix it there first

CJ42 (approved):

PR #391, opened 2024-02-26
fix: Allow LF or TAB in the middle of JSON detection

richtera:

Allow JSON detection to work if there is a LF right before and after the end.


richtera:
@Hugoo This one is required first.

Hugoo (approved):

CJ42 (approved):

PR #394, opened 2024-03-20
feat: add dynamic array encoding for encodeData

fhildeb:
### What kind of change does this PR introduce?

**Feature**: Dynamically encode ERC725 storage keys of type **Array**.

### What is the current behavior?

The library only allows users to define the full array they want to encode and set to a contract. For updates, developers must read through all existing elements and pay the gas for re-setting elements in the list, even if only a few of them are actually updated. This is highly impractical. 

### What is the new behavior?

Within the **encodeData(...)** function, developers can define a **startingIndex** and an **arrayLenght** parameter to dynamically adjust the output, which can be used for more gradient storage updates. Both parameters are optional. If not provided, the encoding falls back to the default behavior as before, starting from index 0 and using the length of the provided elements in the array.

### DIFF
--- a/docs/classes/ERC725.md
+++ b/docs/classes/ERC725.md
@@ -444,13 +444,13 @@ _Example 2: input `1122334455` encoded as `bytes4` --> will encode as `0x42e576f
 
 An array of objects containing the following properties:
 
-| Name                         | Type                                           | Description                                                                                                                                                      |
-| :--------------------------- | :--------------------------------------------- | :--------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| `keyName`                    | string                                         | Can be either the named key (i.e. `LSP3Profile`, `LSP12IssuedAssetsMap:<address>`) or the hashed key (with or without `0x` prefix, i.e. `0x5ef...` or `5ef...`). |
-| `dynamicKeyParts` (optional) | string or <br/> string[&nbsp;]                 | The dynamic parts of the `keyName` that will be used for encoding the key.                                                                                       |
-| `value`                      | string or <br/> string[&nbsp;] <br/> JSON todo | The value that should be encoded. Can be a string, an array of string or a JSON...                                                                               |
-| `startingIndex` (optional)   | number                                         | Starting index for `Array` types to encode a subset of elements. Defaults t `0`.                                                                                 |
-| `arrayLength` (optional)     | number                                         | Parameter for `Array` types, specifying the total length when encoding a subset of elements. Defaults to the number of elements in the `value` field.            |
+| Name                          | Type                                           | Description                                                                                                                                                      |
+| :---------------------------- | :--------------------------------------------- | :--------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `keyName`                     | string                                         | Can be either the named key (i.e. `LSP3Profile`, `LSP12IssuedAssetsMap:<address>`) or the hashed key (with or without `0x` prefix, i.e. `0x5ef...` or `5ef...`). |
+| `dynamicKeyParts` (optional)  | string or <br/> string[&nbsp;]                 | The dynamic parts of the `keyName` that will be used for encoding the key.                                                                                       |
+| `value`                       | string or <br/> string[&nbsp;] <br/> JSON todo | The value that should be encoded. Can be a string, an array of string or a JSON...                                                                               |
+| `startingIndex` (optional)    | number                                         | Starting index for `Array` types to encode a subset of elements. Defaults t `0`.                                                                                 |
+| `totalArrayLength` (optional) | number                                         | Parameter for `Array` types, specifying the total length when encoding a subset of elements. Defaults to the number of elements in the `value` field.            |
 
 The `keyName` also supports dynamic keys for [`Mapping`](https://github.com/lukso-network/LIPs/blob/main/LSPs/LSP-2-ERC725YJSONSchema.md#mapping) and [`MappingWithGrouping`](https://github.com/lukso-network/LIPs/blob/main/LSPs/LSP-2-ERC725YJSONSchema.md#mapping). Therefore, you can use variables in the key name such as `LSP12IssuedAssetsMap:<address>`. In that case, the value should also set the `dynamicKeyParts` property:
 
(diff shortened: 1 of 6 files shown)


Hugoo (changes requested):
Can you also update the function JSDoc at lines 231, 232. It is missing the 2 new inputs.

Can you update the docs and add examples on this new feature?

Nice tests

src/lib/utils.test.ts
```diff
@@ -720,6 +720,168 @@ describe('utils', () => {
...
+    it('should be able to specify the array length + starting index', () => {
+      const encodedArraySection = encodeData(
+        [
+          {
+            keyName: 'AddressPermissions[]',
+            value: [
+              '0x983abc616f2442bab7a917e6bb8660df8b01f3bf',
+              '0x56ecbc104136d00eb37aa0dce60e075f10292d81',
+            ],
+            arrayLength: 23,
```
Hugoo, line 743: I have the feeling this can be a bit flaky. Like ppl can mess up a bit with this.

What about changing this to "currentArrayLength"

And then, the value you return is `currentArrayLength + value.length`

Then there is no way to have a wrong number, or at least it is harder to fail it.


Because right now, if let's say:

- current array of size = 5
- you add 3 values
- you give `arrayLength=4`

You have a valid (as per your code) values.length > arrayLength. But you have a wrong value of array length = 4 but it should be 5 + 3 = 8.

I think this suggestion might make it less error prone.

what do you think?

src/lib/utils.ts
```diff
@@ -260,21 +262,46 @@ export function encodeKey(
         return null;
       }
 
+      // Check if an starting index and array lenght are correct types
```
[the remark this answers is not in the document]
Hugoo: Actually, you can remove this comment. It is so obvious from the code. This comment is not needed.

Less code less problems.

src/lib/utils.test.ts
Hugoo: For next time, I suggest to open issues and have the discussion on the issue page so everyone (incl. me and external Devs) can get the whole discussion 😉

Hugoo: Then maybe you can call it "finalArrayLength"?

Or something. Just as arrayLength can be confusing. Is it before? After? etc.

docs/classes/ERC725.md
```diff
@@ -700,6 +702,50 @@ myErc725.encodeData([
...
+  {
+    name: 'AddressPermissions[]',
+    key: '0xdf30dba06db6a30e65354d9a64c609861f089545ca58c6b4dbe31a5f338cb0e3',
+    keyType: 'Array',
+    valueType: 'address',
+    valueContent: 'Address',
+  },
+];
+
+const myErc725 = new ERC725(schemas);
```
Hugoo, line 719: Please keep this example consistent with the other examples of that function.

1. No instantiation of ERC725 (why would this one example have `new ERC725(...)` and not the other ones? Either all get it or none.
2. You can use it like that:

```js
const schemas = [
  {
    name: 'DynamicKey:<bytes4>:<string>',
    key: '0xForDynamicKeysThisFieldIsIrrelevantAndWillBeOverwriten',
    keyType: 'Mapping',
    valueType: 'bytes',
    valueContent: 'Address',
  },
];

myErc725.encodeData(
  [
    {
      keyName: 'DynamicKey:<bytes4>:<string>',
      dynamicKeyParts: ['0x11223344', 'Summer'],
      value: '0xcafecafecafecafecafecafecafecafecafecafe',
    },
  ],
  schemas,
);
```

fhildeb:
Thanks for the review and feedback 🙏🏻

I renamed `arrayLength` to `totalArrayLength` for better clarification. I also added a new info box and warning regarding the usage of these elements. 

I think it's important to eliminate all questions, as this will be the second "big update" for dynamic user inputs, besides the already existing `dynamicKeyParts`...which also has some extended explanation in the docs. 👍🏻

src/index.test.ts
```diff
@@ -1004,6 +1004,45 @@ describe('Running @erc725/erc725.js tests...', () => {
           assert.deepStrictEqual(results, intendedResult);
         });
 
+        it('encodes dynamic data values for keyType "Array" in naked class instance', () => {
```
CJ42, line 1007: name of test seems incorrect.

src/lib/utils.ts
```diff
@@ -260,21 +262,45 @@ export function encodeKey(
...
+        typeof totalArrayLength !== 'number'
+      ) {
+        throw new Error(
+          'Invalid startingIndex or totalArrayLength parameters. Values must be of type number.',
+        );
+      }
+
+      if (startingIndex < 0) {
+        throw new Error(
+          'Invalid startingIndex parameter. Value cannot be negative.',
```
CJ42, line 276: Suggest showing variable in quote in error string.

```suggestion
          'Invalid `startingIndex` parameter. Value cannot be negative.',
```

src/lib/utils.ts
```diff
@@ -260,21 +262,45 @@ export function encodeKey(
...
+        );
+      }
+
+      if (totalArrayLength < value.length) {
+        throw new Error(
+          'Invalid totalArrayLength parameter. Array length must be at least as large as the number of elements of the value array.',
```
CJ42, line 282: same comment as above

CJ42 (approved):
LGTM

src/lib/utils.ts
fhildeb: I was not able to add brackets or double quotes in Error outputs. Tried both `"`, ```, `\`` due to the linter configuration.

@CJ42 do you have another idea to improve the outputs?

CJ42: Maybe if you escape them with backslash `\`. But it is a minor detail, it is fine as it is.

CJ42 (approved):

src/lib/utils.ts
fhildeb: As you see above, I tried backslash, but the linter removes it automatically. I don't know what's better- ignore formatting and linting for those lines or keep it as is. I would argue the second one is more convenient.

When I edit the error messages, the `throw` tests also start to fail:

img1

Hugoo (approved):

src/lib/utils.ts
richtera: You don't need a backslash for different quotes. What was the error when you put a plain backquote around the startingIndex and totalArrayLength? Usually error message don't contain back quotes, but I don't really see a reason why not.

fhildeb:
Alright, I figured out how to adjust the error messages. Had to adjust the tests to also include these brackets, as they were looking for a close match.

PR #409, opened 2024-04-04
fix: Filter out ALL_PERMISSIONS from keys in LSP6_DEFAULT_PERMISSIONS.

richtera:

Filter out keys from LSP6_DEFAULT_PERMISSIONS because it now has ALL_PERMISSIONS.
Having ALL_PERMISSIONS set during encodePermissions will always return ALL_PERMISSIONS.
Currently calling decodePermissions will always return ALL_PERMISSIONS on.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@erc725/erc725.js",
-  "version": "0.24.0",
+  "version": "0.24.1-dev.5",
   "description": "Library to interact with ERC725 smart contracts",
   "main": "build/main/src/index.js",
   "typings": "build/main/src/index.d.ts",
@@ -16,6 +16,8 @@
     "build:module": "tsc -p tsconfig.module.json",
     "test": "env TS_NODE_COMPILER_OPTIONS='{\"module\": \"commonjs\" }' nyc --reporter=text --reporter=lcov mocha",
     "lint": "eslint . --ext .ts",
+    "format:fix": "prettier --write .",
+    "format": "prettier .",
     "release": "standard-version"
   },
   "repository": {
@@ -51,6 +53,7 @@
   },
   "homepage": "https://github.com/ERC725Alliance/erc725.js#readme",
   "devDependencies": {
+    "@biomejs/biome": "^1.6.4",
     "@types/chai": "^4.3.4",
     "@types/isomorphic-fetch": "^0.0.36",
     "@types/jest": "^27.5.2",
--- a/src/index.test.ts
+++ b/src/index.test.ts
@@ -27,7 +27,14 @@ import { hexToNumber, leftPad, numberToHex } from 'web3-utils';
 // examples of schemas to load (for testing)
 import { LSP1Schema, LSP12Schema, LSP3Schema, LSP6Schema } from './schemas';
 
-import ERC725 from '.';
+import ERC725, {
+  checkPermissions,
+  decodeMappingKey,
+  decodePermissions,
+  encodeKeyName,
+  encodePermissions,
+  supportsInterface,
+} from '.';
 import {
   decodeKeyValue,
   encodeKey,
@@ -988,9 +995,9 @@ describe('Running @erc725/erc725.js tests...', () => {
             values: [],
           };
 
-          keyValuePairs.forEach((keyValuePair) => {
-            intendedResult.keys.push(keyValuePair.key);
-            intendedResult.values.push(keyValuePair.value);
+          keyValuePairs.forEach(({ key, value }) => {
+            intendedResult.keys.push(key);
+            intendedResult.values.push(value);
           });
 
           const erc725 = new ERC725([schemaElement]);
@@ -1394,7 +1401,7 @@ describe('Running @erc725/erc725.js tests...', () => {
       testCases.forEach((testCase) => {
         it(`Encodes ${testCase.hex} permission correctly`, () => {
           assert.deepStrictEqual(
-            ERC725.encodePermissions(testCase.permissions),
+            encodePermissions(testCase.permissions),
             testCase.hex,
           );
           assert.deepStrictEqual(
@@ -1406,7 +1413,7 @@ describe('Running @erc725/erc725.js tests...', () => {
 
       it('Defaults permissions to false if not passed', () => {
         assert.deepStrictEqual(
-          ERC725.encodePermissions({
+          encodePermissions({
             EDITPERMISSIONS: true,
             SETDATA: true,
           }),
@@ -1475,7 +1482,7 @@ describe('Running @erc725/erc725.js tests...', () => {
       for (let i = 0; i < numberOfTests; i++) {
         it(`Randomized test #${i + 1}`, () => {
           const randomPermissions = generateRandomPermissions();
-          const encoded = ERC725.encodePermissions(randomPermissions);
+          const encoded = encodePermissions(randomPermissions);
           const expectedHex = calculateExpectedHex(randomPermissions);
           assert.strictEqual(
             encoded,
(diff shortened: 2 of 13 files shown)


.vscode/settings.json
CJ42, line 1: do we want to have these files?

src/lib/permissions.ts
```diff
@@ -1,60 +1,37 @@
...
-  // Do not use an for of loop here to not require the regenerator runtime
-  for (let i = 0; i < additionalPermissions.length; i += 1) {
-    const permission = additionalPermissions[i];
-    if (permissions[permission]) {
-      basePermissions |= BigInt(
-        hexToNumber(LSP6_DEFAULT_PERMISSIONS[permission]),
-      );
-    }
-  }
+  let basePermissions = BigInt(0)
```
richtera, line 6: This is where the encodePermissions was changed.

src/lib/permissions.ts
```diff
@@ -83,33 +60,27 @@ export function decodePermissions(permissionHex: string) {
     SIGN: false,
```
richtera, line 60: This is where the decodePermissions was changed.

CJ42 (changes requested):
can we remove biome from this PR, and put it in a separate PR please?
Otherwise the PR is really hard to review as there are too many changes and I can't filter out what is related to the permissions

src/index.test.ts
```diff
@@ -1382,170 +1393,223 @@ describe('Running @erc725/erc725.js tests...', () => {
...
           encoded,
           LSP6_DEFAULT_PERMISSIONS.ALL_PERMISSIONS,
-          'Encoded permissions do not match expected value for ALL_PERMISSIONS',
-        );
-      });
+          'Encoded permissions do not match expected value for ALL_PERMISSIONS'
+        )
+      })
+
+      it('should decode ALL_PERMISSIONS', () => {
```
richtera, line 1517: Here is the first new test

src/index.test.ts
```diff
@@ -1382,170 +1393,223 @@ describe('Running @erc725/erc725.js tests...', () => {
...
+        const decodedPermissions = decodePermissions(encoded)
+
+        assert.strictEqual(
+          decodedPermissions.ALL_PERMISSIONS,
+          true,
+          'Decoded permissions includes ALL_PERMISSIONS'
+        )
+      })
+
+      it('should not decode CALL or ALL_PERMISSIONS if perms are missing', () => {
```
richtera, line 1530: Here is the second new test

richtera (commented):
Since there are a lot of style changes here are the real changes.

.vscode/settings.json
richtera: Yes, these files are necessary to make sure any devs use the correct settings. It's kind of unfortunate but currently .editorconfig doesn't properly defined formatters and validators in a generic way. If this were to use a standard lsp https://microsoft.github.io/language-server-protocol/ then any editor could hook into it.

richtera:
I removed import sorting, so now only the cleanup and fixes are there.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@erc725/erc725.js",
-  "version": "0.24.0",
+  "version": "0.24.1-dev.5",
```
CJ42, line 3: why `-dev.5`?

package.json
```diff
@@ -16,6 +16,8 @@
     "build:module": "tsc -p tsconfig.module.json",
     "test": "env TS_NODE_COMPILER_OPTIONS='{\"module\": \"commonjs\" }' nyc --reporter=text --reporter=lcov mocha",
     "lint": "eslint . --ext .ts",
+    "format:fix": "prettier --write .",
```
CJ42, line 19: nice addition 👌

src/index.test.ts
```diff
@@ -1498,12 +1505,81 @@ describe('Running @erc725/erc725.js tests...', () => {
...
+
+        assert.strictEqual(
+          decodedPermissions.ALL_PERMISSIONS,
+          true,
+          'Decoded permissions includes ALL_PERMISSIONS',
+        );
+      });
+
+      it('should not decode CALL or ALL_PERMISSIONS if perms are missing', () => {
+        const permissions = { ALL_PERMISSIONS: true, CALL: false };
```
CJ42, line 1522: @skimaharvey was not the intended logic to be "all permissions except `CALL`" in this kind of case?

src/index.test.ts
```diff
@@ -1498,12 +1505,81 @@ describe('Running @erc725/erc725.js tests...', () => {
...
+        const decodedPermissions = decodePermissions(encoded);
+
+        assert.strictEqual(
+          decodedPermissions.CALL,
+          false,
+          'Decoded permissions includes CALL',
+        );
+        assert.strictEqual(
+          decodedPermissions.ALL_PERMISSIONS,
+          false,
```
CJ42, line 1534: shouldn't this be `true`?

```suggestion
          true,
```

src/index.test.ts
```diff
@@ -1498,12 +1505,81 @@ describe('Running @erc725/erc725.js tests...', () => {
...
+
+        assert.strictEqual(
+          redecodedPermissions.CALL,
+          false,
+          'Re-reencoded permissions includes CALL',
+        );
+      });
+
+      it('should allow editing of permissions add extra', () => {
+        const permissions = { ALL_PERMISSIONS: true };
```
CJ42, line 1556: how about a test like that:

```suggestion
        const permissions = { ALL_PERMISSIONS: true, SUPER_DELEGATECALL: true, DELEGATECALL: true };
```

Should lead to the same result right?

src/lib/permissions.ts
```diff
@@ -87,22 +59,16 @@ export function decodePermissions(permissionHex: string) {
...
-      result[testPermission] = true;
-    }
-    return result;
-  }
-
-  const passedPermissionDecimal = Number(hexToNumber(permissionHex));
+  const passedPermissionDecimal = BigInt(hexToNumber(permissionHex, true));
 
   // Do not use an for of loop here to not require the regenerator runtime
+  // Deal with ALL_PERMISSIONS the same way. So as long as all the bits in ALL_PERMISSIONS
```
CJ42, line 65: I am very confused by this comment. I think might need a bit of re-writing and maybe a small example?

CJ42 (changes requested):
Added some review comments. Otherwise some nice cleanups here! 🙌

src/index.test.ts
```diff
@@ -1498,12 +1505,81 @@ describe('Running @erc725/erc725.js tests...', () => {
...
+
+        const decodedPermissions = decodePermissions(encoded);
+        decodedPermissions.CALL = false;
+        const reencodePermissions = encodePermissions(decodedPermissions);
+        const redecodedPermissions = decodePermissions(reencodePermissions);
+
+        assert.strictEqual(
+          redecodedPermissions.CALL,
+          false,
+          'Re-reencoded permissions includes CALL',
```
CJ42, line 1551: isn't this error message incorrect?

```suggestion
          'Re-reencoded permissions does not include CALL',
```

src/index.test.ts
```diff
@@ -1498,12 +1505,81 @@ describe('Running @erc725/erc725.js tests...', () => {
...
+        decodedPermissions.DELEGATECALL = true;
+        const reencodePermissions = encodePermissions(decodedPermissions);
+        const redecodedPermissions = decodePermissions(reencodePermissions);
+
+        assert.strictEqual(
+          redecodedPermissions.SUPER_DELEGATECALL,
+          true,
+          'Re-reencoded permissions includes SUPER_DELEGATECALL',
+        );
+        assert.strictEqual(
```
CJ42, line 1570: I guess we can also add a test for:

```js
        assert.strictEqual(
          redecodedPermissions.ALL_PERMISSIONS,
          true,
          'Re-reencoded permissions includes ALL_PERMISSIONS',
        );
```

package.json
CJ42: will be overriten by release please, but let's remove it in case.

src/index.test.ts
CJ42: yes that's the case. All good

CJ42 (approved):

PR #428, opened 2024-04-28
feat: add `dynamicName` in schema returned by `getSchema`

CJ42:

:star: Feature

#400 introduced a change in the `getSchema` function that replaces the dynamic part `<...>` with the actual hex value in the `name` of the schema.

However, this causes a problem has the schem returned cannot be used anymore as it will the `name` now contains a hex value and `ERC725.encodeData` or `ERC725.decodeData` will not find the name anymore when searching through the data keys.

- [x] Add an extra property `dynamicName` returned in the schema when dealing with `Mapping` and `MappingWithGrouping` data keys when using `getSchema(...)`
- [x] Preserves the full `name` of the data key to address the issue mentioned above.
- [x] Export `getSchema` function standalone, to make it available outside of class instance

### Other information:

None

### DIFF
--- a/src/lib/schemaParser.test.ts
+++ b/src/lib/schemaParser.test.ts
@@ -13,7 +13,7 @@
 */
 
 import assert from 'assert';
-import { DynamicNameSchema } from '../types/ERC725JSONSchema';
+import { ERC725JSONSchema } from '../types/ERC725JSONSchema';
 
 import { getSchema } from './schemaParser';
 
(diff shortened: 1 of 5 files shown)


src/lib/schemaParser.test.ts
```diff
@@ -131,11 +133,13 @@ describe('schemaParser getSchema', () => {
...
         '1111222233334444555566667777888899990000aaaabbbbccccddddeeeeffff';
-      const name = `SomeBytes32Mapping:${bytes32Value}`;
+      const name = `SomeBytes32Mapping:<bytes32>`;
+      const dynamicName = `SomeBytes32Mapping:${bytes32Value}`;
       const key = `0x0cfc51aec37c55a4d0b10000${bytes32Value.slice(0, 42)}`;
 
-      const extraSchema: ERC725JSONSchema = {
+      const extraSchema: DynamicNameSchema = {
         name,
+        dynamicName,
```
richtera, line 142: Hmmm, this should really be dynamicKeyParts: string | string[]. You're welcome to have dynamicName, but when passing it to encodeKey it will expect dynamicKeyParts, no?

richtera (changes requested):

richtera (approved):

richtera (approved):

richtera (approved):

richtera (approved):

PR #432, opened 2024-05-13
feat: export methods  `getVerificationMethod(..)` and `isDataAuthentic(...)` related to `VerifiableURI` encoding

CJ42:

✨ Feature

Some methods defined in the lib are only available internally, but exposed to the class or exported in the main index.js of the lib.

Add the following methods + their relevant docs:
- `getVerificationMethod`
- `isDataAuthentic`


src/lib/getDataFromExternalSources.test.ts
```diff
@@ -52,4 +52,134 @@ describe('getDataFromExternalSources', () => {
...
+          },
+          url: 'ipfs://QmdMGUxuQsm1U9Qs8oJSn5PfY4B1apGG75YBRxQPybtRVm',
+        },
+      },
+    ];
+
+    const result = await getDataFromExternalSources(
+      [schema],
+      dataFromChain,
+      'https://api.universalprofile.cloud/ipfs/',
```
Hugoo, line 82: Please mock this as we can't rely on outside API/server for the testing of this lib.

Hugoo (changes requested):
I would suggest to consider having a proper mock to remove external dependencies.

Hugoo:
docs updated here: https://github.com/ERC725Alliance/erc725.js/pull/447/files

Hugoo (approved):

PR #436, opened 2024-05-13
feat: map permission and permissions docs section

CJ42:

✨ Feature

The function `mapPermission` can be useful to retrieve the bytes32 hex value of a permission given its name. However, it is not available.

Export the `mapPermission` function to make it available in the library + refactor it to return `null` when the permission name is not known instead of throwing an error.


docs/methods/external-data-source-utilities.md
```diff
@@ -0,0 +1,384 @@
...
+```
+
+```js
+ERC725.isDataAuthentic(data, verificationOptions);
+```
+
+```js
+import { isDataAuthentic } from '@erc725/erc725.js';
+
+isDataAuthentic(data, verificationOptions);
```
richtera, line 221: Note this will not always work for the decoded json data unfortunately. If the checksum was pretty printed then the data needs to be the byte array of the file.

richtera (approved):

Hugoo (commented):
I see the docs are already live for this feature:

https://docs.lukso.tech/tools/erc725js/methods#mappermission

PR #449, opened 2024-05-17
feat!: allow to encode array index using `dynamicKeyPart` in `encodeData` + return `dynamicName` with `getSchema` for Arrays

CJ42:

🌟 Feature

Currently, it is not possible when using `encodeData` to simply encode array index data keys. 

Add feature for arrays that enable to pass a `dynamicKeyPart` as a `number` that will encode for an array index.


src/lib/schemaParser.test.ts
```diff
@@ -56,15 +56,19 @@ describe('schemaParser getSchema', () => {
...
           key: '0x7c8c3416d6cda87cd42c71ea1843df2800000000000000000000000000000001',
+          dynamicName: 'LSP12IssuedAssets[1]',
+          dynamicKeyPart: '0x00000000000000000000000000000001',
           keyType: 'Singleton',
           valueContent: 'Address',
           valueType: 'address',
         },
         '0xdf30dba06db6a30e65354d9a64c6098600000000000000000000000000000000': {
-          name: 'AddressPermissions[0]',
+          name: 'AddressPermissions[]',
```
Hugoo, line 68: This is a breaking change - we need to flag it as such

Hugoo (approved):

PR #451, opened 2024-05-27
fix: allow to decode `0x` as empty array `[]` for keyType `Array`

CJ42:

Bug fix

When value is `0x` to try to `decodeData` for key type `Array`, the following error occurs.

image

Allow to decode `0x` as empty array `[]`.


src/lib/Untitled-1
JeneaVranceanu, line 1: It looks like this file was not supposed to be committed 🙂

JeneaVranceanu (approved):

Hugoo (approved):

PR #459, opened 2024-06-24
feat: Reapply upgrades and add msw

richtera:

fix: Upgrade npm modules, fix test scripts, add some exports, fix LF JSON parsing issue


examples/package.json
```diff
@@ -9,10 +9,10 @@
     "decodeData": "node src/decodeData.js"
   },
   "dependencies": {
-    "@erc725/erc725.js": "^0.17.2",
+    "@erc725/erc725.js": "..",
```
CJ42, line 12: maybe we can just put this?

```suggestion
    "@erc725/erc725.js": "latest",
```

examples/package.json
```diff
@@ -9,10 +9,10 @@
...
     "web3": "^4.7.0"
   },
   "devDependencies": {
-    "isomorphic-fetch": "^3.0.0"
+    "cross-fetch": "^3.0.0"
```
CJ42, line 16: why was this replaced?

examples/src/instantiation.js
```diff
@@ -2,11 +2,11 @@ import Web3 from 'web3';
...
 
 // this is needed because node does not support `fetch` out of the box
-// isomorphic-fetch is not needed in a browser environment
-import 'isomorphic-fetch';
+// cross-fetch is not needed in a browser environment
+import 'cross-fetch';
 
 const RPC_ENDPOINT = 'https://rpc.testnet.lukso.network';
-const IPFS_GATEWAY = 'https://2eff.lukso.dev/ipfs/';
+const IPFS_GATEWAY = 'https://api.universalprofile.cloud/ipfs/';
```
CJ42, line 9: nice catch!

package.json
```diff
@@ -47,38 +55,46 @@
...
   },
   "dependencies": {
-    "add": "^2.0.6",
-    "ethereumjs-util": "^7.1.5",
-    "web3-eth-abi": "^1.10.0",
-    "web3-providers-http": "^1.10.0",
-    "web3-utils": "^1.10.0"
+    "cross-fetch": "^4.0.0",
+    "global-agent": "^3.0.0",
+    "msw": "^2.3.1",
```
CJ42, line 87: why do we have these top 3 now in `"dependencies"`? Is it necessary for the lib to run?
Asking as the library is small, we we want to keep to a minimum and wondering if it will increase the size.

package.json
```diff
@@ -47,38 +55,46 @@
...
+    "web3": "^4.5.0",
+    "web3-eth-abi": "^4.2.0",
+    "web3-eth-accounts": "^4.1.1",
+    "web3-providers-http": "^4.1.0",
+    "web3-utils": "^4.1.1",
+    "web3-validator": "^2.0.4"
+  },
+  "ts-node": {
```
CJ42, line 95: For my knowledge, why is this new field needed?

src/index.test.ts
```diff
@@ -655,38 +654,30 @@ describe('Running @erc725/erc725.js tests...', () => {
...
-            ),
-          );
           const result = await erc725.fetchData('TestJSONURL');
           assert.deepStrictEqual(result, {
             key: testJSONURLSchema.key,
             name: testJSONURLSchema.name,
             value: JSON.parse(jsonString),
           });
-        } finally {
-          fetchStub.restore();
```
CJ42, line 689: looks like this `finally` block is gone. Is that normal?

src/index.test.ts
```diff
@@ -702,38 +693,19 @@ describe('Running @erc725/erc725.js tests...', () => {
           [contractVersion.interface],
         );
 
-        const ipfsGateway = 'https://api.universalprofile.cloud';
```
CJ42, line 705: nice cleanup! 🧹

src/index.ts
```diff
@@ -84,8 +83,9 @@ import {
 } from './lib/permissions';
 import { AssetURLEncode } from './types/encodeData';
 import { URLDataToEncode, URLDataWithHash, Verification } from './types';
+import { isAddress } from 'web3-validator';
 
-export {
+export type {
```
CJ42, line 88: For my knowledge, why is the `type` keyword now needed?

src/lib/_test.ts
CJ42, line 1: What is this new file for?

src/lib/encodeKeyName.ts
```diff
@@ -112,24 +107,23 @@ export const encodeDynamicKeyPart = (
...
-      if (valueWithoutPrefix.length !== size * 2) {
+      if (!isHex(value)) {
         throw new Error(
-          `Wrong value: ${value} for dynamic key with type: ${type}. Value is not ${size} bytes long.`,
+          `Wrong value: ${value} for dynamic key with type: ${type}. Value is not in hex.`,
         );
       }
-
-      if (!isHex(valueWithoutPrefix)) {
+      const valueWithoutPrefix = value.replace('0x', '');
```
CJ42, line 115: Suggest using library function
https://docs.web3js.org/api/web3-eth-accounts/function/stripHexPrefix

```suggestion
      const valueWithoutPrefix = stripHexPrefix(value);
```

CJ42 (commented):
Added review comments

examples/package.json
richtera: Ok we could do that, I have .. so it installs the currently compiled code.

package.json
richtera: Hmm I forget "global-agent", but "cross-fetch" is because of the data-providers so we don't include fetch two times. Checking on global-agent.

richtera: Removing global-agent. This is a thing for debugging similar to fiddler on windows. It allows you to enable proxies for apps so you can watch the traffic.

examples/package.json
CJ42: ok we might need to make a mention of this in our docs.

src/index.test.ts
```diff
@@ -2012,7 +1972,19 @@ describe('decodeMappingKey', () => {
...
     assert.deepStrictEqual(
-      erc725Instance.decodeMappingKey(
+      decodeMappingKey(
+        encodeKeyName('MyKeyName:<bytes16>', ['0x12345678']),
+        'MyKeyName:<bytes16>',
+      ),
+      [
+        {
+          type: 'bytes16',
+          value: '00000000000000000000000012345678',
```
CJ42, line 1982: This is incorrect, `bytesN` are right padded, so it should be:

```suggestion
          value: '12345678000000000000000000000000',
```

src/provider/providerWrapper.test.ts
```diff
@@ -53,6 +53,7 @@ describe('ProviderWrapper', () => {
       try {
         await ethSource.getOwner(erc725AccountAddress);
       } catch (error: any) {
+        // console.log(error);
```
CJ42, line 56: This might have been forgotten to be removed maybe?

src/provider/providerWrapper.ts
```diff
@@ -98,9 +92,10 @@ export class ProviderWrapper {
       ERC725Y_INTERFACE_IDS.legacy,
     );
 
-    return isErc725YLegacy
-      ? ERC725_VERSION.ERC725_LEGACY
-      : ERC725_VERSION.NOT_ERC725;
+    if (isErc725YLegacy) {
```
CJ42, line 95: Nice prefer this syntax too

CJ42 (changes requested):

examples/package.json
CJ42: https://github.com/lukso-network/docs/issues/1065

Opened issue for now

src/index.test.ts
CJ42: > the spy stuff doesn't work in ESM, that's why all the finally stub things are gone

src/index.test.ts
richtera: I tried to fix various things, but it really went down the rabbit hole. I would like to talk on Monday. I created a new PR which shows where I found/fixed problems.

richtera (commented):
I made a follow up PR to talk about this. It seems large and it pulled in required changes to make all the parse decode runnable.

CJ42 (approved):

PR #460, opened 2024-06-25
docs: add docs examples for encoding / decoding tuples

CJ42:


src/lib/decodeData.test.ts
```diff
@@ -164,7 +171,7 @@ describe('decodeData', () => {
     expect(decodedData.name).to.eql('KeyOne');
   });
 
-  it('parses type tuples/Mixed correctly', () => {
+  it.only('parses type tuples/Mixed correctly', () => {
```
Hugoo, line 174: need to remove the only here

src/lib/decodeData.test.ts
```diff
@@ -185,6 +192,8 @@ describe('decodeData', () => {
       [schema],
     );
 
+    console.log('decodedData for tuple: ', decodedData);
```
Hugoo, line 195: to remove :)

PR #496, opened 2025-02-10
Added browser IDEs

mojafa:
Hey hey,

Mo here. I've been diving into open source lately and thought I'd give my mojo a little boost. I added a few browser IDEs to the readme, hoping it might make things a bit easier for new contributors getting started with the codebase.

What do you think? Hope it resonates!

Best,
Mo.


richtera:
@mojafa I like this, but maybe the buttons should be specific per example or rather have a single one up to to pull the whole repo. What do you think?

CJ42:
We need to decide as we already have Stackblitz and we also use CodeSandbox in docs.lukso.tech
So need to pick one and stick to it IMO

PR #497, opened 2025-03-07
improvement: fetchData return null if no key not found

doubleppereira:
improvement to fetchData method
right now fetchData throws an exception if we try to fetchData of a non existing key. We think it should return just null in this case
if fetchData receives an non existing key on contract it should just return null

### DIFF
--- a/.tool-versions
+++ b/.tool-versions
@@ -1,1 +1,1 @@
-nodejs 20.x
+nodejs 20.0.0
--- a/biome.json
+++ b/biome.json
@@ -23,7 +23,7 @@
     "formatter": {
       "semicolons": "always",
       "quoteStyle": "single",
-      "trailingComma": "all"
+      "trailingCommas": "all"
     }
   },
   "organizeImports": {
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@erc725/erc725.js",
-  "version": "0.24.0",
+  "version": "0.27.1",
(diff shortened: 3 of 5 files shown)


richtera:
When fetching a single key the response is an error, when fetching multiple then it will return null and not throw an error. The idea is that if you ask for a specific key then it's ok to get an error and you can deal with it.

richtera:
I added support for this by default. So if there is an error it will show up as a null value with an error property that you can read if you need it. That way it will not throw errors by default, but they are still available @CJ42 FYI.

richtera:
Added inside of https://github.com/ERC725Alliance/erc725.js/pull/500

- (here1)[https://github.com/ERC725Alliance/erc725.js/blob/8ca81141fa75a8995e1c32cb3f4d2ebed9b81df5/src/lib/getDataFromExternalSources.ts#L171]
- (here2)[https://github.com/ERC725Alliance/erc725.js/blob/8ca81141fa75a8995e1c32cb3f4d2ebed9b81df5/src/lib/getDataFromExternalSources.ts#L178]
- (here3)[https://github.com/ERC725Alliance/erc725.js/blob/8ca81141fa75a8995e1c32cb3f4d2ebed9b81df5/src/lib/getDataFromExternalSources.ts#L186]

Basically it will return the error as an error property on the output data rather than throwing it.

richtera:
This is now included in #500

CJ42:
Closing as included in #504

Issue #499, opened 2025-04-17
Bug in encodeTupleKeyValue


tehnalogos:
Example of bug

`encodeTupleKeyValue("(Bytes4,Number)", "(bytes4,uint256)", ["0xc52d6008", 1])`

expected result
`"0xc52d60080000000000000000000000000000000000000000000000000000000000000001"`
actual result
`"0xc52d60080000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000001"`

richtera:
Adding it into the test script on my PR.

richtera:
Added
- ✔ encodes/decodes keyType Array / tuples (valueContent: (Bytes4,Number), valueType: (bytes4,uint256)
Test case
```js
{
        schema: {
          name: 'TupleWithBytes4Uint256',
          key: '0x4b80742de2bf393a64c70000<address>',
          keyType: 'Singleton',
          valueType: '(bytes4,uint256)',
          valueContent: '(Bytes4,Number)',
        },
        valueType: '',
        decodedValue: ['0xc52d6008', 1] as string[] | Array<string | string[]>,
        encodedValue:
          '0xc52d60080000000000000000000000000000000000000000000000000000000000000001',
      }
```

CJ42:
Closing as completed in #499 by @richtera

PR #504, opened 2025-05-01
fix: Support for arrays in tuples. Testscript fixes, improved coverage (to develop)

richtera:

- fix: Add array support inside of tuples
- fix: [CompactBytesArray] type of arrays only work at the end of a tuple
- fix: Make all test scripts pass, use real msw, test ethers and viem.
- fix: Change to version 20 and 22 of node.
- fix: Add testing using legacy RPC provider supporting send(payload, callback) in addition to supporting async send(payload). We don't currently use async send(method, params) at all.
- fix: Add test case to cover https://github.com/ERC725Alliance/erc725.js/issues/499
- fix: Allow external spec for fetch/convert
- fix: Repair vulnerabilities
- fix: Cleanup old eslint and mov msw to be properly in devDeps. Also do some upgrades and npm audit checks.
- fix: Found some more testscripts missing and unexpected schema concat (provided needs to be first.)
- fix: Disable codecov badge since we don't currently have it.
- fix: Improve some coverage

- feat: Allow passing of custom fetch and convertUrl implementation for IPFS gateway.
- feat: Allow enabling exceptions when validating schema.
- feat: msw RPC simulation during test scripts.

- Currently when the library sees [CompactBytesArray] as a string it assumes it's an array of tuples, then it removes the string and treats the array element inside of the tuple as it's plain component type.

- Allows usage of abi encoded arrays inside of tuples as well as [CompactBytesArray] type of arrays as the last element of the tuple.


.github/workflows/lint-test-build.yml
```diff
@@ -37,7 +41,19 @@ jobs:
         run: npm run build --if-present
 
       - name: Test
-        run: npm test
+        if: matrix.node-version == '22.x'
```
CJ42, line 44: Why testing only with nodejs v22?

package.json
```diff
@@ -54,42 +59,39 @@
...
+    "sinon": "^20.0.0",
     "standard-version": "^9.5.0",
     "ts-node": "^10.9.2",
-    "typescript": "^5.3.3"
+    "typescript": "^5.8.3",
+    "viem": "^2.28.1"
   },
   "dependencies": {
-    "cross-fetch": "^4.0.0",
-    "msw": "^2.3.1",
```
CJ42, line 86: Is there a particular reason why `msw` is gone?

reporter-config.json
```diff
@@ -0,0 +1,6 @@
+{
+  "reporterEnabled": "spec, json",
```
CJ42, line 2: is there not a dot missing here? Just wondering.

```suggestion
  "reporterEnabled": ".spec, .json",
```

src/lib/decodeData.test.ts
```diff
@@ -70,6 +70,20 @@ describe('decodeData', () => {
...
+      key: '0x6de85eaf5d982b4e5da00000<address>',
+      keyType: 'Mapping',
+      valueType: '(', // BAD
+      valueContent: '(Bytes4,Number)',
+    },
+    {
+      name: 'LSP4CreatorsMap:<address>',
+      key: '0x6de85eaf5d982b4e5da00000<address>',
+      keyType: 'Mapping',
+      valueType: 'bytes4,uint128', // BAD2
```
CJ42, line 84: nice new tests 👌🏻

src/lib/getData.ts
```diff
@@ -41,7 +41,7 @@ const getArrayValues = async (
 
   const arrayLength = await decodeKeyValue(
     'Number',
-    'uint128',
+    'uint128+',
```
CJ42, line 44: why is there a `+` sign here? This must be a typo

```suggestion
    'uint128',
```

src/lib/getDataFromExternalSources.test.ts
```diff
@@ -53,6 +53,28 @@ describe('getDataFromExternalSources', () => {
         IPFS_GATEWAY_MOCK,
       );
     }).to.not.throw();
+    expect(async () => {
+      await getDataFromExternalSources(
+        [
+          {
+            name: 'LSP3Profile2',
```
CJ42, line 60: Why the `2`?

package.json
CJ42: ah yes it was moved to dev dependencies, since it is for mocking. Make sense!

src/lib/utils.test.ts
```diff
@@ -224,19 +286,123 @@ describe('utils', () => {
         encodedValue:
           '0x002000000003ca41e4ea94c8fa99889c8ea2c8948768cbaf4bc03e89ad98ffffffff002000000002f70ce3b58f275a4c28d06c98615760dde774de57ffffffff760d9bba002000000001d3236aa1b8a4dde5ea375fd1f2fb5c354e686c9fffffffffffffffff',
       },
+      {
+        schema: {
+          name: 'TupleWithArray',
+          key: '0x4b80742de2bf393a64c70000<address>',
+          keyType: 'MappingWithGrouping',
+          valueType: '(bytes4,bytes8,bytes32[CompactBytesArray])',
```
CJ42, line 294: Shouldn't this be `uint64`? Since the `valueContent` is `Number`

```suggestion
          valueType: '(bytes4,uint64,bytes32[CompactBytesArray])',
```

src/lib/utils.test.ts
```diff
@@ -224,19 +286,123 @@ describe('utils', () => {
...
+      {
+        schema: {
+          name: 'TupleWithArrayPartialMore',
+          key: '0x4b80742de2bf393a64c70000<address>',
+          keyType: 'MappingWithGrouping',
+          valueType: '(bytes4,bytes8,bytes32[CompactBytesArray])',
+          valueContent: '(Bytes4,Number,Bytes32)',
+        },
+        valueType: '',
+        decodedValue: ['0xdeadbeaf', 12, null] as
```
CJ42, line 360: Couldn't we improve the typing for the provided inputs so we don't have to use `as`?

src/lib/utils.test.ts
```diff
@@ -224,19 +286,123 @@ describe('utils', () => {
...
+          | Array<string | string[]>,
+        encodedValue: '0xdeadbeaf000000000000000c',
+        encodedError: '0xdeadbeaf000000000000000c002001234342343', // This encoded value has the length and partial byte of the encoded array instead of null
+      },
+      {
+        schema: {
+          name: 'TupleWithArray',
+          key: '0x4b80742de2bf393a64c70000<address>',
+          keyType: 'MappingWithGrouping',
+          valueType: '(bytes4,bytes32[],bytes8)',
```
CJ42, line 371: same here

```suggestion
          valueType: '(bytes4,bytes32[],uint64)',
```

src/lib/utils.test.ts
```diff
@@ -224,19 +286,123 @@ describe('utils', () => {
...
+        decodedValue: [
+          '0xdeadbeaf',
+          [
+            '0x1234567812345678123456781234567812345678123456781234567812345678',
+            '0x2345678123456781234567812345678123456781234567812345678123456789',
+          ],
+          12,
+        ] as string[] | Array<string | string[]>,
+        encodedValue:
+          '0xdeadbeaf0000000000000000000000000000000000000000000000000000000000000020000000000000000000000000000000000000000000000000000000000000000212345678123456781234567812345678123456781234567812345678123456782345678123456781234567812345678123456781234567812345678123456789000000000000000c',
```
CJ42, line 384: perfect 👌🏻

CJ42 (approved):
added review comments

.github/workflows/lint-test-build.yml
richtera: We test with both but we only put the 22 report into the PR. For the other one (20) the test results end up inside of the build log.

reporter-config.json
richtera: No those are names of reporters not file extensions

src/lib/getData.ts
richtera: This means that we can read all the bytes to allow the old uint255 as well as the mistake uint16. When we see the plus then we read all the hex bytes and calculate how big the uintN should be. This is only used for the array length item

src/lib/getDataFromExternalSources.test.ts
richtera: To make it non-existent or non matching the key

src/lib/utils.test.ts
richtera: I added some typedefs.

.github/workflows/lint-test-build.yml
richtera: So for v22 we run `npm run test:ci` and for 20 we run `npm run test`

PR #506, opened 2025-05-07
feat: Upgrade to viem and bigint

richtera:

Upgrade to viem and bigint for the main implementation.

### DIFF
--- a/src/constants/constants.ts
+++ b/src/constants/constants.ts
@@ -12,7 +12,7 @@
     along with @erc725/erc725.js.  If not, see <http://www.gnu.org/licenses/>.
 */
 
-import { numberToHex, keccak256, toHex } from 'web3-utils';
+import { numberToHex, keccak256, toHex, toBytes } from 'viem';
 
 import { type MethodData, Encoding, Method } from '../types/Method';
 
@@ -113,7 +113,7 @@ export const SUPPORTED_VERIFICATION_METHODS_LIST = Object.values(
 
 export function keccak256Method(data: object | string | Uint8Array | null) {
   if (data === null) {
-    return keccak256('');
+    return keccak256(new Uint8Array([]));
   }
   if (data instanceof Uint8Array) {
     const buffer = toHex(data);
@@ -121,9 +121,9 @@ export function keccak256Method(data: object | string | Uint8Array | null) {
   }
   if (typeof data === 'object') {
     const buffer = JSON.stringify(data);
-    return keccak256(buffer);
+    return keccak256(toBytes(buffer));
   }
-  return keccak256(data);
+  return keccak256(toBytes(data));
 }
 
 const KECCAK256_UTF8 = {
--- a/src/index.ts
+++ b/src/index.ts
@@ -115,7 +115,7 @@ export {
   mapPermission,
 } from './lib/permissions';
 export { getSchema } from './lib/schemaParser';
-import { createPublicClient, http, isAddress } from 'viem';
+import { createPublicClient, Hex, http, isAddress } from 'viem';
 import { lukso } from 'viem/chains';
 
 // PRIVATE FUNCTION
@@ -522,8 +522,8 @@ export class ERC725 {
    * @returns true if isValidSignature call on the contract returns the magic value. false otherwise
    */
   async isValidSignature(
-    messageOrHash: string,
-    signature: string,
+    messageOrHash: Hex | string,
+    signature: Hex,
   ): Promise<boolean> {
     if (!this.options.address || !isAddress(this.options.address)) {
       throw new Error('Missing ERC725 contract address.');
--- a/src/index.ts
+++ b/src/index.ts
@@ -127,7 +127,7 @@ function initializeProvider(providerOrRpcUrl, gasInfo) {
   if (typeof providerOrRpcUrl === 'string') {
     const client = createPublicClient({
       chain: lukso,
-      transport: http('https://rpc.testnet.lukso.network'),
+      transport: http(providerOrRpcUrl),
     });
     return new ProviderWrapper(client, gasInfo);
   }
--- a/src/index.ts
+++ b/src/index.ts
@@ -21,8 +21,6 @@
  * @date 2020
  */
 
-import HttpProvider from 'web3-providers-http';
-
 import { ProviderWrapper } from './provider/providerWrapper';
 
 import {
@@ -82,7 +80,6 @@ import {
 } from './lib/permissions';
 import type { AssetURLEncode } from './types/encodeData';
 import type { URLDataToEncode, URLDataWithHash, Verification } from './types';
-import { isAddress } from 'web3-validator';
 
 export type {
   ERC725JSONSchema,
@@ -118,6 +115,8 @@ export {
   mapPermission,
 } from './lib/permissions';
 export { getSchema } from './lib/schemaParser';
+import { createPublicClient, http, isAddress } from 'viem';
+import { lukso } from 'viem/chains';
 
 // PRIVATE FUNCTION
 function initializeProvider(providerOrRpcUrl, gasInfo) {
@@ -126,7 +125,11 @@ function initializeProvider(providerOrRpcUrl, gasInfo) {
 
   // if provider is a string, assume it's a rpcUrl
   if (typeof providerOrRpcUrl === 'string') {
-    return new ProviderWrapper(new HttpProvider(providerOrRpcUrl), gasInfo);
+    const client = createPublicClient({
+      chain: lukso,
+      transport: http('https://rpc.testnet.lukso.network'),
+    });
+    return new ProviderWrapper(client, gasInfo);
   }
 
   if (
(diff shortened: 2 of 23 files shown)


src/index.ts
```diff
@@ -126,7 +125,11 @@ function initializeProvider(providerOrRpcUrl, gasInfo) {
 
   // if provider is a string, assume it's a rpcUrl
   if (typeof providerOrRpcUrl === 'string') {
-    return new ProviderWrapper(new HttpProvider(providerOrRpcUrl), gasInfo);
+    const client = createPublicClient({
+      chain: lukso,
+      transport: http('https://rpc.testnet.lukso.network'),
```
doubleppereira, line 130: should we be hardcoding the rpc url here? and is it going to be always testnet?

richtera: Ah good catch, thanks.

src/constants/constants.ts
```diff
@@ -113,17 +113,17 @@ export const SUPPORTED_VERIFICATION_METHODS_LIST = Object.values(
 
 export function keccak256Method(data: object | string | Uint8Array | null) {
   if (data === null) {
-    return keccak256('');
+    return keccak256(new Uint8Array([]));
```
CJ42, line 116: super!

src/index.ts
```diff
@@ -126,7 +125,11 @@ function initializeProvider(providerOrRpcUrl, gasInfo) {
 
   // if provider is a string, assume it's a rpcUrl
   if (typeof providerOrRpcUrl === 'string') {
-    return new ProviderWrapper(new HttpProvider(providerOrRpcUrl), gasInfo);
+    const client = createPublicClient({
+      chain: lukso,
```
CJ42, line 129: hmmmm but if I want to use that on LUKSO Testnet, then it would not work because you are passing mainnet as `chain`?

src/lib/schemaParser.ts
```diff
@@ -88,7 +88,10 @@ const fillDynamicKeyPart = (
 
     // if first 20 bytes of the hash of second word in schema match,
     // display the map part as plain word
-    if (keccak256(keyNameParts[1]).substring(0, 42) === `0x${secondWordHex}`) {
+    if (
+      keccak256(toBytes(keyNameParts[1])).substring(0, 42) ===
```
CJ42, line 92: Couldn't we use a hex slicing function from viem? It would be cleaner here over `substring`

src/lib/encoder.ts
```diff
@@ -82,20 +80,22 @@ export const encodeDataSourceWithHash = (
...
-      verification?.data
-        ? padLeft(verification.data.slice(2).length / 2, 4)
-        : '0000',
-    ),
-    stripHexPrefix(
-      verification?.data ? stripHexPrefix(verification?.data) : '',
-    ),
-    stripHexPrefix(utf8ToHex(dataSource)),
+    (verificationMethod
+      ? pad(keccak256(toBytes(verificationMethod.name)).slice(0, 10) as Hex, {
```
CJ42, line 84: Suggest using the `slice` function from viem here

```suggestion
      ? pad(slice(keccak256(toBytes(verificationMethod.name)), 0, 4), {
```

https://viem.sh/docs/utilities/slice

image

CJ42: Also this line is very big and hard to read:

```ts
const verificationNameAsUtf8Hex = toBytes(verificationMethod.name);

const verificationMethodId = slice(keccak256(verificationNameAsUtf8Hex), 0, 4));
```

CJ42: I am also not sure why here we are padding on the left. Why is padding needed? Since we are taking just the first 4 bytes of the hash digest

src/lib/encoder.ts
```diff
@@ -212,33 +212,50 @@ type BytesNValueTypes =
...
-  if (typeof value === 'string' && !isHex(value)) {
+  if (!isHex(value)) {
     // if we receive a plain string (e.g: "hey!"), convert it to utf8-hex data
     valueToEncode = toHex(value);
   } else if (typeof value === 'number') {
     // if we receive a number as input, convert it to hex,
     // despite `bytesN` pads on the right, we pad number on the left side here
     // to symmetrically encode / decode
-    valueToEncode = padLeft(numberToHex(value), numberOfBytesInType * 2);
+    valueToEncode = pad(numberToHex(value), {
```
CJ42, line 228: This looks so much better over the `* 2` from web3.js

src/lib/encoder.ts
```diff
@@ -367,7 +387,7 @@ const decodeBytesNCompactBytesArray = (
 ): string[] => {
   const bytesValues = decodeCompactBytesArray(compactBytesArray, consumed);
   bytesValues.forEach((bytesValue, index) => {
-    if (stripHexPrefix(bytesValue).length > numberOfBytes * 2)
+    if (bytesValue.slice(2).length > numberOfBytes * 2)
```
CJ42, line 390: I think since there is no `stripHexPrefix` function in viem, we could create our own local utility function in here? That would make more sense codewise when reading these `<...>.slice(2)` functions

src/types/decodeData.ts
```diff
@@ -3,8 +3,8 @@ import type { EncodeDataType, URLDataWithHash } from './encodeData/JSONURL';
 
 export interface DataInput {
   keyName: string; // can be the name or the hex/hash
-  value;
-  dynamicKeyParts?: string | string[] | number;
+  value: unknown;
```
CJ42, line 6: Why `unknown`?

src/provider/providerWrapper.ts
```diff
@@ -135,9 +134,9 @@ export class ProviderWrapper {
     signature: string,
   ): Promise<string> {
     if (this.type === ProviderTypes.ETHEREUM) {
-      const encodedParams = encodeParameters(
-        ['bytes32', 'bytes'],
-        [hash, signature],
+      const encodedParams = encodeAbiParameters(
+        [{ type: 'bytes32' }, { type: 'bytes' }],
+        [hash as Hex, signature as Hex],
```
CJ42, line 139: Can't we replace the function param type to `0x${string}` as well here? (or `Hex`) so that we can improve the typing and remove all of these `as` inside the function?

```ts
  async isValidSignature(
    address: string,
    hash: Hex,
    signature: Hex,
```

CJ42: There are some good types in viems that can be imported. But not sure how much it would be breaking for applications.

https://viem.sh/docs/glossary/types.html#hash

image

src/lib/utils.ts
```diff
@@ -670,12 +684,16 @@ export function patchIPFSUrlsIfApplicable(
...
+    return 0;
+  }
+
+  return data.slice(2).length / 2;
 }
 
-export function countSignificantBits(data: string) {
-  const bytes = hexToBytes(data);
+export function countSignificantBits(data: Hex) {
+  const bytes = hexToBytes(data as Hex);
```
CJ42, line 696: This shouldn't be necessary, since it is the parameter type no?

```suggestion
  const bytes = hexToBytes(data);
```

src/lib/utils.ts
```diff
@@ -565,7 +576,10 @@ export function hashData(
   data: string | Uint8Array | Record<string, any>,
   nameOrSig: SUPPORTED_VERIFICATION_METHODS | string,
 ): string {
-  return getVerificationMethod(nameOrSig)?.method(data) || leftPad(0, 64);
+  return (
+    getVerificationMethod(nameOrSig)?.method(data) ||
+    pad(toHex(0), { size: 32, dir: 'left' })
```
CJ42, line 581: Suggest using this function

```suggestion
    numberToHex(0, { size: 32 })
```

image

src/lib/utils.ts
```diff
@@ -170,7 +171,10 @@ export function encodeKeyValue(
  * @return The raw bytes key for the array element
  */
 export function encodeArrayKey(key: string, index: number) {
-  return key.slice(0, 34) + padLeft(numberToHex(index), 32).replace('0x', '');
+  return (
+    key.slice(0, 34) +
```
CJ42, line 175: Same here, suggest using the `slice(data, size)` function from viem library

CJ42: Could also use the `concat` method from viem here

image

CJ42 (changes requested):
This looks much better with viem! I have added some review comments

src/index.ts
richtera: The configured public client could provide the chain, but this is just the wrapper to allow specification of the URL. It will connect to testnet just fine, the chain is mainly used to get the URL (which is specificed) the multicall contracts (which is correct but we're not using anyways) and the explorer which we're not using in this scenario.

src/lib/schemaParser.ts
richtera: Yes we could for sure. Checking into it.

src/lib/encoder.ts
richtera: The padding was for 0x0 but you're right most of the time it's 0x00000000.
Fixed.

src/types/decodeData.ts
richtera: Unknown because you don't know whether it's an object or Hex or a number. That way you don't have to do the as any thing.

src/lib/utils.ts
richtera: True

CJ42 (approved):

